PR #249, opened 2014-12-10
Fixed bugs in path calculations

Ihromant:
Fixed bugs in path calculations

@mcm69 or @shamansir please review and test whether it works.


shamansir:
Will merge it after merging PR #256

PR #256, opened 2015-01-08
Linter fixes

mcm69:


src/anm/animation/element.js
```diff
@@ -1211,11 +1245,12 @@ Element.prototype.removeModifier = function(modifier) {
     if (!this.__modifiers_hash[modifier.id]) throw new AnimationError('Modifier wasn\'t applied to this element');
     if (!modifier.__applied_to || !modifier.__applied_to[this.id]) throw new AnimationError(Errors.A.MODIFIER_NOT_ATTACHED);
     //if (this.__modifying) throw new AnimErr("Can't remove modifiers while modifying");
-    delete this.__modifiers_hash[modifier.id];
-    delete this.$modifiers[modifier.type].splice(modifier.__applied_to[this.id] - 1, 1); // delete by index
-    delete modifier.__applied_to[this.id];
+    utils.removeElement(this.__modifiers_hash, modifier.id);
```
shamansir: `removeElement` name correlates with DOMElement or Animatron Element, I thinks it's more like `disposeValue` or something.

src/anm/events.js
```diff
@@ -79,21 +79,20 @@ function provideEvents(subj, events) {
...
+    };
+
     for (var ei = 0, el = events.length; ei < el; ei++) {
-        _event = events[ei];
-        subj.prototype['e_'+_event] = (function(event) {
-            return function(evtobj) {
-                this.fire(event, evtobj);
-            };
-        })(_event);
+        subj.prototype['e_'+_event] = makeFireFunc(events[ei]);
```
shamansir: Where's `_event` value assigned here?

src/anm/events.js
```diff
@@ -79,21 +79,20 @@ function provideEvents(subj, events) {
         for (var evt in _hdls) {
             if (_hdls.hasOwnProperty(evt)) _hdls[evt] = [];
         }
-    }
+    };
     /* FIXME: call fire/e_-funcs only from inside of their providers, */
     /* TODO: wrap them with event objects */
-    var _event;
+    var makeFireFunc = function(event){
```
shamansir: I vote for spaces before the curly braces :)

src/anm/player.js
```diff
@@ -24,7 +24,7 @@ var C = require('./constants.js'),
     conf = require('./conf.js'),
     log = require('./log.js'),
     is = utils.is,
-    events = require('./events.js');
+    events = require('./events.js'),
```
shamansir: oh, here you are, bastard! :+1:

src/anm/events.js
mcm69: good catch, thanks

src/anm/resource_manager.js
```diff
@@ -168,13 +174,13 @@ ResourceManager.prototype.check = function() {
             if (errors[urls[u]]) error_count++;
```
shamansir: please move `u` definition from this loop in `var` block above it then, similarly to the cases before.

src/anm/utils.js
```diff
@@ -199,6 +197,13 @@ function fit_rects(pw, ph, aw, ah) {
     } else return [ 1, [ 0, 0, aw, ah ] ];
 }
 
+function removeElement(array, element) {
```
shamansir: Hmm, it wasn't the array operation at all. It was a method to [delete properties in the object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete).

src/anm/render.js
```diff
@@ -67,14 +67,15 @@ function r_loop(ctx, player, anim, before, after, before_render, after_render) {
...
 }
+
 function r_at(time, dt, ctx, anim, width, height, zoom, rib_color, before, after) {
     ctx.save();
     var ratio = engine.PX_RATIO;
     if (ratio !== 1) ctx.scale(ratio, ratio);
-    var width = width | 0,
-        height = height | 0;
+    width = width | 0;
+    height = height | 0;
```
shamansir: Note to myself: some comment that this is a rounding and not defaulting here should be added.

src/anm/animation/element.js
shamansir: Second thought: `disposeValue` is also not a "true", I only do it to:

* some algorithms are based on property-check here and there, so if, for example, the modifier ID is not in modifiers hash _object_, modifier is treated as removed; I believe these algorithms will break in given case;
* break references so if user has no pointers to it, it won't take the memory;

src/anm/utils.js
mcm69: deleting properties is bad, don't you remember the JS objects presentation?

shamansir: I thought [it's quite ok](http://bonsaiden.github.io/JavaScript-Garden/#core.delete) for [your own objects with no prototype chains](http://perfectionkills.com/understanding-delete/). Anyway, as noted above, some algorithms based on it, so please at least change it to `object[prop] = null`;

shamansir: Ok, [I found why it's bad](http://www.smashingmagazine.com/2012/11/05/writing-fast-memory-efficient-javascript/), thank you, so yeah, `null`-ing is the proper way here (since we need to get `if (modifers[id])` fail to `false`).

shamansir:
Thank you!

PR #261, opened 2015-01-19
Fix shadow scaling

shamansir:
@skavish @mcm69 Please review

### DIFF
--- a/src/anm/player.js
+++ b/src/anm/player.js
@@ -529,9 +529,10 @@ Player.prototype.play = function(from, speed, stopAfter) {
         player.reportStats();
     }
 
+    var ctx_props = engine.getAnmProps(player.ctx);
+    ctx_props.factor = this.factor();
+
     state.happens = C.PLAYING;
-
-
 
     // FIXME: W3C says to call stopAnim (cancelAnimationFrame) with ID
     //        of the last call of nextFrame (requestAnimationFrame),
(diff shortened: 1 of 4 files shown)


src/anm/player.js
```diff
@@ -916,6 +921,48 @@ Player.prototype.size = function(width, height) {
 };
 
 /**
+ * @method factor
+ *
+ * Returns the difference factor between player size and animation size,
+ * using fit by largest side. _Does not_ counts scene zoom, since it does not
```
mcm69: does not _count_

src/anm/render.js
```diff
@@ -89,8 +89,8 @@ function r_at(time, dt, ctx, anim, width, height, zoom, rib_color, before, after
         } finally { ctx.restore(); }
     } else {
         r_with_ribbons(ctx, width, height,
-                              anim.width, anim.height,
-                              rib_color,
+                            anim.width, anim.height,
+                            rib_color,
```
mcm69: why not kill the linebreaks here altogether?

mcm69:
looks good

skavish:
lgtm

src/anm/render.js
shamansir: I like grouping things that visually group together, it reads better and feels better than having long lines of arguments.

src/anm/player.js
shamansir: Will fix, thanks!

PR #262, opened 2015-01-21
PLAY-609 If user clicks in the progress bar near to the end of animation, player fires an error

mcm69:


src/anm/ui/controls.js
```diff
@@ -235,6 +235,12 @@ Controls.prototype.react = function() {
     if (Controls.isInProgressArea(coords, w, h)) {
         if (coords.x > btnWidth && coords.x < w-btnWidth) {
             time = Math.round(p.state.duration*(coords.x-btnWidth)/(w-2*btnWidth));
+            if (time > p.anim.duration) {
+                //when the animation is something like 3.8 seconds long,
+                //the rounding will exceed the duration, which is not
+                //a good idea.
+                time = p.anim.duration;
```
shamansir: I see, we round to seconds, and it is actually visible issue, especially in short clips like 2sec long, when you click, for example, somewhere in 3/4 of progress bar and it jumps to 1sec. I think we should round to hundreds of milliseconds or even less to make it look smoother. `utils.js` has `roundTo` for that cause.

mcm69: for whatever reason I thought `player.play()` would only take a whole number as a time value - i'll have this fixed, thanks!

PR #263, opened 2015-01-22
Fix audio-master issues

shamansir:
Due to removed `_audio_master ` flag, Master-Audio element was put inside Scene element, which was not correct and Scene band was less than Master-band. No flag directly in the Element anymore, moved it to `elm.$audio.master`. Also, renamed `elm.audio` to `elm.$audio` to match other types of elements.

### DIFF
--- a/src/anm/ui/controls.js
+++ b/src/anm/ui/controls.js
@@ -242,7 +242,8 @@ Controls.prototype.react = function() {
                 time = p.anim.duration;
             }
             if (s === C.PLAYING) {
-              p.pause().play(time);
+              p.pause()
+               .play(time);
             } else {
               p.play(time).pause();
             }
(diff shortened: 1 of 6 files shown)


shamansir:
@mcm69 Please review

src/anm/ui/controls.js
```diff
@@ -242,7 +242,8 @@ Controls.prototype.react = function() {
                 time = p.anim.duration;
             }
             if (s === C.PLAYING) {
-              p.pause().play(time);
+              p.pause()
+               .play(time);
```
shamansir: this line-break was made just for easier debugging, providing an ability to set breakpoint to `.pause` :)

mcm69:
looks good!

PR #264, opened 2015-01-26
do not interpolate static color tweens

mcm69:


mcm69:
@shamansir please review

src/anm/utils.js
```diff
@@ -43,6 +43,41 @@ is.tween = function(f) {
...
+    // if both x and y are null or undefined and exactly the same
+
+    if (!(x instanceof Object) || !(y instanceof Object)) return false;
+    // if they are not strictly equal, they both need to be Objects
+
+    if (x.constructor !== y.constructor) return false;
+    // they must have the exact same prototype chain, the closest we can do is
+    // test their constructor.
+
+    for (var p in x) {
```
shamansir: Isn't test for equality becomes more resource-wasting than finding a value between the same ones? Fortunately, most of the player classes have no deep prototype chains. I'd honestly prefer some hash-driven comparison or something like that.

src/anm/graphics/brush.js
```diff
@@ -406,9 +406,18 @@ Brush.clearShadow = function(ctx) {
...
  */
 Brush.interpolateBrushes = function(from, to) {
+    var equal = is.equal(from, to);
     from = (from instanceof Brush) ? from : Brush.value(from);
-    to   = (to   instanceof Brush) ? to   : Brush.value(to);
     if (!from._converted) { from.convertColorsToRgba(); }
+    if (equal) {
+        //if the values are the same, we can just skip the interpolating
+        //and return the first value
+        return function() {
```
shamansir: Also I think there's too much of comments here and there, please do not comment obvious things, there should be a balance :)

src/anm/utils.js
shamansir: For the case of brushes I think it will be much faster if we'll assign some unique integer ID to every new instance and just test these IDs for equality.

mcm69: in case of brushes the comparing is only done once when creating the tween, so I believe it's not a big deal. and we might need this in future, hence a generic equality function

shamansir: Agree with _not a big deal_. But actually I try to fallback to ID-comparison everywhere where it's needed just because I don't really like deep-object analysis :). Object-equality is really not so often-required operation, as for me. For Elements or Modifiers/Painters, where it's needed in several cases, ID's are enough, and do work well.

skavish: what is id-comparison? if it's comparison by reference then it won't work in this case.

shamansir: Nope, it's GUID-comparison, every Element or Modifier has it's short GUID. But yeah, but my statement was wrong, sorry, of course in case of different instances having the same values of Brushes ID-comparison won't work.

PR #266, opened 2015-01-28
Pattern fill and stroke

mcm69:


mcm69:
@skavish @shamansir please review

src/anm/graphics/brush.js
```diff
@@ -176,6 +176,21 @@ Brush.prototype.adapt = function(ctx) {
...
+            fill;
+        var canvas = engine.createCanvas(this.pattern.w, this.pattern.h, null, 1);
+        var cctx = canvas.getContext('2d');
+        elm.pivot(0,0);
+        elm.disabled = false;
+        elm.render(cctx, 0, 0);
+        elm.disabled = true;
+        fill = canvas;
+
+        if (fill) {
```
shamansir: Why `fill` could be `false` here?

src/import/animatron-importer.js
```diff
@@ -744,8 +748,16 @@ Import.fill = function(src) {
...
+    var fill;
+    if (is.str(src[0])) {
+        fill = src[0];
+    } else if (is.arr(src[0])) {
+        if (is.arr(src[0][0])) {
+            fill = Import.grad(src[0]);
+        }
+        fill = Import.pattern(src[0]);
+    }
+    return Brush.stroke(fill, // paint
```
shamansir: In removed lines above, first argument for `Brush.stroke` was `src[1]` for both array-case and non-array-case. Why was it changed to `src[0]` here, while `src[0]` is a width, as the line below states?

src/anm/graphics/brush.js
mcm69: it could be in the previous version, thanks for noticing!

src/import/animatron-importer.js
mcm69: ¯\_(ツ)_/¯

shamansir: I mean, may be export format was changed or it was a bug here before, but then the line below looks improper.

mcm69: I mean "my bad"  ¯\_(ツ)_/¯

shamansir: Uh, ok then :)

shamansir:
lgtm now.

PR #267, opened 2015-01-29
change ordering of resource load

mcm69:
When loading a movie for a second time, the resource manager notified the player that the resources are ready before the actual `loadOrGet`s happened. This led to the player starting playback when some of the elements (namely audio) were not yet ready. This is now fixed.

### DIFF
--- a/src/anm/player.js
+++ b/src/anm/player.js
@@ -377,6 +377,8 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
             state.happens = C.RES_LOADING;
             player.fire(C.S_CHANGE_STATE, C.RES_LOADING);
             player.fire(C.S_RES_LOAD, remotes);
+            // actually start loading remote resources
+            anim._loadRemoteResources(player);
             // subscribe to wait until remote resources will be ready or failed
             resourceManager.subscribe(player.id, remotes, [ player.__defAsyncSafe(
                 function(res_results, err_count) {
@@ -400,8 +402,6 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
                     }
                 }
             ) ]);
-            // actually start loading remote resources
-            anim._loadRemoteResources(player);
         }
 
     };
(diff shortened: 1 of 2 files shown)


mcm69:
@shamansir please review

src/anm/player.js
```diff
@@ -377,6 +377,8 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
             state.happens = C.RES_LOADING;
             player.fire(C.S_CHANGE_STATE, C.RES_LOADING);
             player.fire(C.S_RES_LOAD, remotes);
+            // actually start loading remote resources
+            anim._loadRemoteResources(player);
```
shamansir: Hard to say, honestly. If it's not breaking something or fixing something, then let it be. Though, by logic, it seems to me that subscription should be performed _before_ the actual process of loading, so subscriber will get the result even if some resource was received asynchronously faster than the next call of `subscribe` was performed, which is zero-chance and paranoia, yes, but makes feel safer in the order it was before.

mcm69: The added `check()` is exactly for the case you described - if the resource is already available, be it cache or super-fast loading, then we just go on.

An added bonus is that the movie now starts instantaneously when loaded the second time - it still took some loader spinning before as the RM went through X cycles of `check()` where X is the number of remote resources (and `check()` isn't really the simplest operation as it is).

skavish:
if it works as described then it's awesome!

src/anm/player.js
shamansir: I look in this case now and this way we get a separate new subscribers to an url, because RM logic is designed like: 

* Someone's subscribes to a bunch resources
* All resources trigger to start loading
* If they some of them are in cache, just give them directly to subscribers. And if they are in cache, no `check()` method should be called, see:

https://github.com/Animatron/player/blob/develop/src/anm/resource_manager.js#L98
https://github.com/Animatron/player/blob/develop/src/anm/resource_manager.js#L139

But in the new case it goes like:
* Start loading all resources (loading is started synchronously, of course, then for most of the cases goes asynchronous)
* Also, in loaders, a separate new subscriber is created for every URL, since no-one subscribed to it for now, and `me._waiting[url]` is `null` [here](https://github.com/Animatron/player/blob/develop/src/anm/resource_manager.js#L102), so it jumps to [this]
(https://github.com/Animatron/player/blob/develop/src/anm/resource_manager.js#L121) branch of code.
* Subscription is performed (after resources started loading).

I think if `check()` was a bottleneck, then it should be optimized, but subscribing after triggering the load process looks illogical for me.

I feel I'd have to revert it back for now, since it breaks the RM logic for me.

PR #269, opened 2015-02-03
make the utils.fit_rects return values whole

mcm69:
@shamansir this looks like a safe fix to avoid a nasty IE10 bug (https://connect.microsoft.com/IE/feedback/details/772169/canvas-clearrect-fails-if-clip-rectangle-has-non-integer) when rendering w/ ribbons, please take a look


shamansir:
Looks valid, thank you!

PR #270, opened 2015-02-10
Show Resource Loading Completeness

shamansir:
Draws a tiny 2px-line at the bottom of controls to show the completeness of resources loading process. 

For audio, it has the ability to monitor chunks, [this demo](http://jspro.brothercake.com/media-events/progress.html) and [this article](http://www.sitepoint.com/essential-audio-and-video-events-for-html5/) were used to find out how to do it. For images, it just reports when they are done loading. The total number of URLs requested is used as `1.0`, each file is then `1 / <number-of-urls>` part in the progress line. For audio files, this value is then divided into sub-parts.

Demo

(sorry, I was able to catch only the last moments of loading :) ).

Needs testing.

Also I ask to merge this just after the request, since it contains ResourceManager changes which should be considered by Yuri in his proposals and/or changes :). Code review fixes could be done with the next pull requests.

@alexeypegov @mcm69 please review.

### DIFF
--- a/src/anm/graphics/sheet.js
+++ b/src/anm/graphics/sheet.js
@@ -61,7 +61,7 @@ Sheet.prototype.load = function(player_id, callback, errback) {
         return;
     }
     resMan.loadOrGet(player_id, me.src,
-        function(notify_success, notify_error) { // loader
+        function(notify_success, notify_error, notify_progress) { // loader
             var src = me.src;
             if (https) {
                 src = src.replace('http:', 'https:');
--- a/src/anm/media/audio.js
+++ b/src/anm/media/audio.js
@@ -141,6 +141,8 @@ Audio.prototype.load = function(player) {
                     el.removeEventListener("progress", progressListener, false);
                     el.removeEventListener("canplay", canPlayListener, false);
                     notify_success(el);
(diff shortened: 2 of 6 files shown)


mcm69:
Please use the circle and draw loading progress there instead of introducing a new design element.

shamansir:
There was a discussion somewhere between issues or chats where we all stated out that this should be a line like this, at the bottom, because circle doesn't work very well for this purpose, it's not easy for user to get what's actually happening.

mcm69:
how can you not get a circle getting filed?
also, i'd vote for going along with Artem's design and add the percentage in the center.

shamansir:
Ok, let's move it to Slack, I though everything was clear. We haven't the idea of filling circle though.

alexeypegov:
personally, I think that loading progress should be similar to playing one

On Wednesday, February 11, 2015, Yuriy Opryshko <notifications@github.com>
wrote:

> Please use the circle and draw loading progress there instead of
> introducing a new design element.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Animatron/player/pull/270#issuecomment-73859044>.
>


-- 

--
PEGOV

PR #271, opened 2015-02-23
Basic Video Support

shamansir:
Plays video in a scene previously imported from Animatron, applies tweens, reacts on pause. Wasn't tested for cross-browserability ftm.


shamansir:
@alexeypegov @skavish @mcm69 you're welcome to review

src/anm/media/video.js
```diff
@@ -0,0 +1,153 @@
...
+            if (engine.isHttps) { url = url.replace('http:', 'https:'); }
+
+            var el = engine.createVideo();
+            el.setAttribute("preload", "auto");
+            el.style.display = 'none';
+
+            var progressListener = function(e) {
+                var buffered = el.buffered;
+                if (buffered.length == 1) {
+                    if (el.readyState === 4) {
```
skavish: what is 4?

shamansir: [`HAVE_ENOUGH_DATA`](https://developer.mozilla.org/en/docs/Web/API/HTMLMediaElement). Yeah, may be it should be moved to constant or a have a comment there or even be a part of enum or..

src/anm/media/audio.js
```diff
@@ -349,5 +329,48 @@ Audio.prototype.connect = function(element) {
...
+ */
+Audio.prototype.clone = function() {
+    var clone = new Audio('');
+    clone.url = this.url;
+    clone.offset = this.offset;
+    return clone;
+};
+
+// workaround, see http://stackoverflow.com/questions/10365335/decodeaudiodata-returning-a-null-error
+function syncStream(node){
```
skavish: just curious if you already experienced this problem yourself or this code is "just in case"?

skavish:
lgtm

src/anm/media/audio.js
shamansir: I just moved @alexeypegov's code from top to bottom in this case :). I just prefer to place class code above the util functions, it was the reason.

PR #272, opened 2015-02-27
consider audio playable when readystate=3

mcm69:
IE actually only updates with `readyState = 3` when loading audios from cache, so this is necessary.


shamansir:
@alexeypegov please take a look

PR #273, opened 2015-03-02
PLAY-590 Player should stop playing when tab is not visible

mcm69:


mcm69:
@shamansir plz review

src/anm/player_manager.js
```diff
@@ -40,4 +45,23 @@ PlayerManager.prototype.getPlayer = function(cvs_id) {
...
+            player.pause();
+        } else if (!hidden && player._pausedViaHidden) {
+            player._pausedViaHidden = false;
+            player.play(player.state.from);
+        }
+    }
+};
+
+var manager = new PlayerManager();
+engine.onDocumentHiddenChange = function(hidden) {
```
shamansir: I think modifying something in the engine from outside is not a very good way, could it be refactored like this?:

`engine.onDocumentHiddenChange(function(hidden) { manager.handleDocumentHiddenChange(hidden); });`, which in its turn calls `addEventListener` with corresponding event, if it's there, or does nothing if it is not supported?

mcm69: sounds good to me

src/anm/player_manager.js
```diff
@@ -40,4 +45,23 @@ PlayerManager.prototype.getPlayer = function(cvs_id) {
     return this.hash[cvs_id];
 };
 
-module.exports = new PlayerManager();
+PlayerManager.prototype.handleDocumentHiddenChange = function(hidden) {
```
shamansir: please add docs that this method is private.

src/engine/dom-engine.js
```diff
@@ -981,5 +981,34 @@ $DE.isLocal = local;
...
+} else if (typeof document.mozHidden !== "undefined") {
+  hidden = "mozHidden";
+  visibilityChange = "mozvisibilitychange";
+} else if (typeof document.msHidden !== "undefined") {
+  hidden = "msHidden";
+  visibilityChange = "msvisibilitychange";
+} else if (typeof document.webkitHidden !== "undefined") {
+  hidden = "webkitHidden";
+  visibilityChange = "webkitvisibilitychange";
+}
```
RobertoUa: so much code duplicated. Why not just prepend ['webkit', 'ms', 'moz'] to 'Hidden' and 'visibilitychange'.
Sorry, I'm just DRY fanboy.

PR #274, opened 2015-03-11
Avoid try catch

shamansir:
Several optimizations suggested by Chrome Profiler when tested with large numbers of objects, should highly improve performance:

* `try/catch` blocks were removed from render loop, the error system is based on events and passing the error to the subject before firing it now.
* excessive events like `X_DRAW` were removed, ability to handle `Player` events was removed from `Element` and `Animation` code, now only a subject of event may catch it's event, to avoid loops going too deep with no reason (keyboard/mouse events will still be passed to element's children if they were in bounds of element); the latter could cause errors with an audio/video, they rely on some Player events like `S_START`/`S_STOP`, so it _could be dangerous_;
* Excessive "safe" methods were removed because they used `try`/`catch` to handle errors;

Please **do not merge** until it will be safe to deploy.

### DIFF
--- a/Jakefile
+++ b/Jakefile
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2014 by Animatron.
+ * Copyright (c) 2011-2015 by Animatron.
  * All rights are reserved.
  *
  * Animatron player is licensed under the MIT License, see LICENSE.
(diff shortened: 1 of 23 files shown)


skavish:
can't wait to see it in action!

mcm69:
how about we throw it on test and see what happens?

shamansir:
Yeah, let's do it. Latest commit in develop is 2964e9bc18 for the records.

mcm69:
@shamansir There are still issues, so please in the future announce you're about to merge a PR so I could at least take a quick look.

shamansir:
@mcm69 

> how about we throw it on test and see what happens?

I suppose it means "to merge into `develop` and deploy", no? 'Cause `develop` should match player at test, and `master` should match player at production

mcm69:
i meant 'deploy it on test and let someone take a look while keeping the PR open', i am sorry if I was unclear.

mcm69:
in case we enforce the develop/test and prod/master rule, we'll need something like skunkworks website to test bleeding edge player there

PR #275, opened 2015-03-12
Controls: Remove a loading circle in center, make loading bar higher

shamansir:


src/anm/player.js
```diff
@@ -385,6 +386,7 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
                     if (player.anim === result) { // avoid race condition when there were two requests
                         // to load different animations and first one finished loading
                         // after the second one
+                        player.playedOnce = false;
```
mcm69: why not assign `playedOnce` outside the condition before even bothering with remotes?

src/anm/ui/controls_theme.json
```diff
@@ -25,12 +25,10 @@
...
     },
     "loading": {
-      "activeColor": "white",
-      "inactiveColor": "rgba(255,255,255,0.5)",
-      "factorBackColor": "rgba(100,100,100,1)",
-      "factorDoneColor": "rgba(255,255,255,1)",
-      "factorErrorColor": "rgba(255,0,0,0.8)",
-      "factorLineWidth": 4
+        "factorBackColor": "rgba(0,0,0,0.7)",
+        "factorDoneColor": "rgba(255,255,255,.8)",
```
mcm69: please change opacity to `0.8` for consistency

shamansir: will fix that, thanks!

src/anm/player.js
shamansir: good point, I had some special reason but since I cannot recall details, it's invalid. Anyway, this flag was intended to be used for end-screen, which I had no chance to finish, so I will just clean things up.

PR #276, opened 2015-03-16
Fix clips containing audio/video

mcm69:
I'm still not sure if the audios would stop after their scene is complete, but at least they _work_ now.

@skavish @shamansir please review but do not merge yet

### DIFF
--- a/src/anm/events.js
+++ b/src/anm/events.js
@@ -142,6 +142,7 @@ registerEvent('X_STOP', 'stop', 'x_stop');
 registerEvent('X_ERROR', 'error', 'x_error');
 
 // * playing (player state)
+registerEvent('S_CHANGE_STATE', 'player_change_state', 'player_change_state');
 registerEvent('S_PLAY', 'play', 'play');
 registerEvent('S_PAUSE', 'pause', 'pause');
 registerEvent('S_STOP', 'stop', 'stop');
--- a/src/anm/events.js
+++ b/src/anm/events.js
@@ -128,6 +128,11 @@ registerEvent('X_STOP', 'bandstop');
 // * Animation or Element error
 registerEvent('X_ERROR', 'error');
 
+// * animation start/stop
+registerEvent('A_START', 'animationstart');
+registerEvent('A_STOP', 'animationstop');
+registerEvent('A_PAUSE', 'animationpause');
+
 // * playing (player state)
 
 registerEvent('S_CHANGE_STATE', 'statechange');
--- a/src/anm/media/audio.js
+++ b/src/anm/media/audio.js
@@ -321,19 +321,21 @@ Audio.prototype.toggleMute = function() {
     }
 };
 /** @private @method connect */
-Audio.prototype.connect = function(element) {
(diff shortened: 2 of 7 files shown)


src/anm/media/audio.js
```diff
@@ -321,19 +321,21 @@ Audio.prototype.toggleMute = function() {
     }
 };
 /** @private @method connect */
-Audio.prototype.connect = function(element) {
+Audio.prototype.connect = function(player) {
     var me = this;
-    element.on(C.X_START, function() {
+    player.on(C.S_PLAY, function() {
```
shamansir: I thought on changing these events to be fired from `Animation` class. Like `ANIM_START`, `ANIM_END`, similar to those (`X_START`/`X_END`) in the element, this way it won't be required to pass player here.

Also, `X_START`/`X_END` shouldn't be removed here and in `video.js`, they're the events fired when the Element lifetime (timeband) starts or ends (not a player starts/ends), Audio & Video should both be stopped in these cases.

src/anm/player.js
```diff
@@ -437,7 +438,18 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
...
+
+/**
+ * Connect audio/video elements with player events
+ */
+var initMedia = function(player) {
+    player.anim.traverse(function(el) {
+        if (el.type === C.ET_AUDIO || el.type === C.ET_VIDEO) {
+            (el.$audio || el.$video).connect(player);
+        }
+    });
```
shamansir: This way it should disconnect/unsubscribe on `player.detach`, I suppose.

src/anm/events.js
```diff
@@ -142,6 +142,7 @@ registerEvent('X_STOP', 'stop', 'x_stop');
 registerEvent('X_ERROR', 'error', 'x_error');
 
 // * playing (player state)
+registerEvent('S_CHANGE_STATE', 'player_change_state', 'player_change_state');
```
shamansir: All player state events are marked as `S_`, I think no need in `player_` prefix, and I hope in the nearby future we'll find some more accurate naming pattern, of course.

And yeah, I forgot to register it when introduced it, thanks for finding that!

src/anm/player.js
shamansir: And it should also unsubscribe on `S_LOAD` event (new animation loaded into Player), by the way. If `S_START`/`S_END` events will fire from `Animation` class instead of `Player`, this won't be required, since `player.anim` is replaced on load.

src/anm/media/audio.js
shamansir: Please consider this code: https://github.com/Animatron/player/blob/2964e9bc1833137b07204d23ac0d292c6219ab0b/src/anm/media/audio.js#L323 `C.X_START`/`C.X_STOP` are Element band events, `C.S_START`/`C.S_STOP` are Player (or Animation band) events, all four are required.

shamansir:
I've added the fixes I've meant, but haven't tested it hard. Will deploy to test later and check how it works.

PR #277, opened 2015-03-20
Interactivity is back, in a better way.

shamansir:
@mcm69 @skavish Please review.

Now, it's winding tests for events, proper bounds tests etc.

P.S. Events are named in a beautiful way now, no requirement in creepy IDs anymore.

### DIFF
--- a/src/anm/animation/element.js
+++ b/src/anm/animation/element.js
@@ -166,7 +166,6 @@ Element.DEFAULT_LEN = Infinity;
 Element._customImporters = [];
 provideEvents(Element, [ C.X_MCLICK, C.X_MDCLICK, C.X_MUP, C.X_MDOWN,
                          C.X_MMOVE, C.X_MOVER, C.X_MOUT,
-                         C.X_KPRESS, C.X_KUP, C.X_KDOWN,
                          C.X_START, C.X_STOP ]);
 /**
  * @method is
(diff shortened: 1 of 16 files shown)


src/anm/animation/element.js
```diff
@@ -1920,10 +1865,10 @@ Element.prototype.offset = function() {
 };
 
 /*Element.prototype.local = function(pt) {
-    this.matrix.transformPoint();
+    return this.matrix.transformPoint(pt);
 }
 Element.prototype.global = function(pt) {
-    this.matrix.transformPoint();
+    return this.matrix.adaptPoint(pt);
```
mcm69: why is this code commented? can we delete it?

src/anm/events.js
```diff
@@ -36,10 +30,10 @@ function provideEvents(subj, events) {
     })(events);
     subj.prototype.on = function(event, handler) {
         if (!this.handlers) throw errors.system('Instance is not initialized with handlers, call __initHandlers in its constructor');
-        if (!this.provides(event)) throw errors.system('Event \'' + C.__enmap[event] +
-                                                 '\' not provided by ' + this);
+        if (!this.provides(event)) throw errors.system('Event \'' + event +
+                                                 '\' is not provided by ' + this);
         if (!handler) throw errors.system('You are trying to assign ' +
-                                    'undefined handler for event ' + event);
+                                    'undefined handler for an event \'' + event + '\'');
```
mcm69: i'd do just `if(!handler) return;`

src/anm/graphics/sheet.js
```diff
@@ -171,6 +171,22 @@ Sheet.prototype.bounds = function() {
...
+ * @param {Object} pt point to check
+ * @param {Number} pt.x
+ * @param {Number} pt.y
+ * @return {Boolean} is point inside
+ */
+ Sheet.prototype.inside = function(pt) {
+    return true; // if point is inside of the bounds, point is considered to be
+                 // inside the image shape
+};
+/**
```
mcm69: you shall have to explain this to me in more detail, i'm afraid i don't get what this is for

src/anm/loader.js
```diff
@@ -61,7 +61,7 @@ Loader.loadAnimation = function(player, anim, callback) {
     if (player.anim) player.anim.dispose();
     // add debug rendering
     if (player.debug && !global_opts.liveDebug) {
-        anim.visitElems(function(e) {e.addDebugRender();}); /* FIXME: ensure not to add twice */
+        anim.traverse(function(e) {e.addDebugRender();}); /* FIXME: ensure not to add twice */
```
mcm69: re: fixme, we can just set a flag on the element in `addDebugRender`.

mcm69:
the code looks good to me. does it work? :trollface:

src/anm/events.js
shamansir: Yeah, ok.

src/anm/animation/element.js
shamansir: It reminds me to add these methods finally, I just haven't finally invented the best/proper way to do it, but yeah.

src/anm/graphics/sheet.js
shamansir: All the graphics (`Path`, `Sheet`, `Text`) instances have `.inside` method from now on. For `Sheet`, in future, (when we re-fix Sprites support) it may return `false`, if this point is not located in a current region of an image. For `Text`, _may be_, it should return `false` if the given point is, i.e. between the letters, but probably it will never be done. Both these cases are enhancements, so for now they always return `true`.

And only a `Path` may return `false` for a point if it's not inside the "surface" of a curve (40% of the code in this PR :)). 

All in all, this method is for second-check if mouse event belongs to the `Element` or not, when the point is for sure already inside of a bounds. At some point it could also be improved also for `Text` and `Sheet`, but not now :).

src/anm/loader.js
shamansir: This `FIXME` could be removed, I realized it's for sure added just once.

shamansir:
Sure! At least partly)

PR #279, opened 2015-03-25
Several fixes for event handling, temporary disable Element.m_on

shamansir:


src/import/animatron-importer.js
```diff
@@ -358,10 +358,11 @@ Import.branch = function(type, src, all, anim) {
         }
 
         if (lsrc[8]) { // scripting
+            anim.hasScripting = true;
             var code = lsrc[8];
             var script_ctx = createScriptContext(ltrg);
 
-            eval('(function(script) {' +
+            eval('(function(script) {'
```
RobertoUa: oh, no

RobertoUa: `SyntaxError: Unexpected identifier`

shamansir: Thanks!

PR #280, opened 2015-03-27
Use path2d

mcm69:
see http://jsperf.com/anm-path-vs-path2d for free performance gains.

### DIFF
--- a/src/anm/graphics/path.js
+++ b/src/anm/graphics/path.js
@@ -8,7 +8,8 @@ var segments = require('./segments.js'),
     LSeg = segments.LSeg,
     CSeg = segments.CSeg,
     Crossings = segments.Crossings,
-    engine = require('engine');
+    engine = require('engine'),
+    useP2D = !!engine.Path2D;
 
 var Brush = require('./brush.js');
 
@@ -59,14 +60,10 @@ function Path(val) {
 
     if (is.str(val)) {
         this.parse(val);
+        this.updatePath2D(val);
     } else if (is.arr(val)) {
         this.segs = val;
     }
-
-    if (engine.Path2D) {
-        this.path2d = new Path2D(Path.toSVGString(this));
-    }
-
     this.cached_hits = {};
 }
 
@@ -120,6 +117,7 @@ Path.prototype.length = function() {
  */
 Path.prototype.add = function(seg) {
     this.segs.push(seg);
+    this._p2dCurrent = false;
     return this;
 };
 /**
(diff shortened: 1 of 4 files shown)


mcm69:
do not merge yet please - i found a bug in SVG parsing when using `new Path(<svg path>)`, I'd like to fix it as a part of this.

shamansir:
I think it only could be merged when it will also provide `bounds`, `hitAt`, `tangentAt` etc. methods using Path2D from the inside, otherwise, Translate/Orient-To-Path tweens and handling events for scripting will all be broken when Path2D will be used inside.

mcm69:
how and _why_ should we do this given that Path2D is a _drawing_ primitive which contains no math and we use it just to _draw_ stuff?

I can see this will fail to respond to path modification, but that is easily fixed - I'll add it to the next commit.

mcm69:
@shamansir does any code actually modify the path itself? I'm not sure.

shamansir:
@mcm69 Yes, you are right, sorry! I somehow thought it _replaces_ the segment data inside with Path2D instance while it's not, for sure. Sorry! It's ok to merge.

src/anm/graphics/path.js
```diff
@@ -580,20 +541,38 @@ var encodeVisitor = function(segment, buffer) {
 Path.parse = function(path, target) {
     target = target || new Path();
     target.segs = [];
-    Path.visitStrPath(path, parserVisitor, target);
+    var segPaths = path.match(/[a-z][^a-z]*/ig);
```
shamansir: Is this way faster? Just asking — it is not very important in general, since path parsing only performed on import. Btw you may store `/[a-z][^a-z]*` in a `RegEx` object and re-use it like a constant.

src/anm/graphics/path.js
```diff
@@ -62,6 +63,10 @@ function Path(val) {
         this.segs = val;
     }
 
+    if (engine.Path2D) {
+        this.path2d = new Path2D(Path.toSVGString(this));
```
skavish: we already have this path parsed, so why convert it to svg string again? i believe it will be faster (and less memory) to create the path using the Path2D api.

src/anm/graphics/path.js
skavish: fast import is also important!

src/anm/graphics/path.js
mcm69: most likely the path will be created from importer (which takes binary path data from the json) - so the SVG string will not be available here. we can rig the importer to convert binary path to svg path though - this way we can drop this line and some others too.

src/anm/graphics/path.js
mcm69: it may not be _way faster_, but the nice thing about it is that it works for any svg path according to the spec (previously, having two spaces or a comma would make the parsing fail).

shamansir:
so... if it's ok to merge, then merge!

mcm69:
just let me fiddle with importer a bit and we should be good

mcm69:
should be good now

PR #281, opened 2015-04-03
Use AWS sdk for deployment

mcm69:


mcm69:
to use `jake invalidate` please add CloudFront distribution id to `.s3`.

PR #283, opened 2015-04-07
Support player.js (http://playerjs.io)

mcm69:
@skavish @shamansir plz review

### DIFF
--- a/src/anm/interop/playerjs-io.js
+++ b/src/anm/interop/playerjs-io.js
@@ -1,0 +1,173 @@
+var engine = require('engine'),
+    C = require('../constants.js');
+
+var playerJsVersion = '0.0.10';
+
+var adapter = {
+    play: function() {
+        this.play();
+    },
+
+    pause: function() {
+        this.pause();
+    },
+
+    isPaused: function() {
+        return this.state.happens === C.PAUSED;
+    },
+
+    mute: function() {
+        this.mute();
+    },
+
+    unmute: function() {
+        this.unmute();
+    },
+
+    isMuted: function() {
+        return this.muted;
+    },
+
+    setVolume: function(message) {
+        var volume = message.value;
+    },
+
+    getVolume: function() {
+        return 100;
+    },
+
+    seekTo: function(message) {
+        var time = message.value;
+        this.pause().play(time);
+    },
+
+    getCurrentTime: function() {
+        return this.state.time;
+    },
+
+    setLoop: function(message) {
+        var loop = message.value;
+        this.repeat = loop;
+    },
+
+    getLoop: function() {
+        return this.repeat;
+    },
+
+    addEventListener: function(message) {
+        var event = message.value,
+            listener = message.listener;
+        if (!listeners[event] || listeners[event].indexOf(listener) !== -1) {
+            return;
+        }
+        listeners[event].push(listener);
+    },
+
+    removeEventListener: function(message) {
+        var event = message.value,
+            listener = message.listener;
+        if (!listeners[event]) {
+            return;
+        }
+        if (!listener) {
+            listeners[event] = [];
+        } else {
+            var index = listeners[event].indexOf(listener);
+            if (index !== -1) {
+                listeners.event.splice(index, 1);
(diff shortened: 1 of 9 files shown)


mcm69:
I'll add `player.seek(time)` method tomorrow, as player.play(time) flashes controls which isn't good if the player is controlled externally, so plz do not merge until then.

shamansir:
`player.seek` would be cool, useful for interactivity where seeking could be performed very often!

src/anm/interop/playerjs-io.js
```diff
@@ -0,0 +1,190 @@
...
+        var loop = message.value;
+        this.repeat = loop;
+    },
+
+    getLoop: function() {
+        return this.repeat;
+    },
+
+    addEventListener: function(message) {
+        var event = message.value,
```
shamansir: Is it possible to use `provideEvents` from `events.js` here instead of custom Event System?

src/anm/player.js
```diff
@@ -380,6 +380,7 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
                         player.state.happens = C.LOADING;
                         player.fire(C.S_CHANGE_STATE, C.LOADING);
                         player.fire(C.S_LOAD, result);
+                        player._updateAudioVolumes();
```
shamansir: Could it be a listener for `S_LOAD` event in `audio.js` module?

src/anm/render.js
```diff
@@ -15,16 +15,21 @@ var fit_rects = require('./utils.js').fit_rects;
...
 
+//time counters to emit S_TIME_UPDATE every second or so
+var timeCounters = {}, TIME_UPDATE_TIME = 0.5;
+
 // draws current state of animation on canvas and postpones to call itself for
 // the next time period (so to start animation, you just need to call it once
 // when the first time must occur and it will chain its own calls automatically)
 function r_loop(ctx, player, anim, before, after, before_render, after_render) {
-
+    if (typeof timeCounters[player.id] === 'undefined') {
```
shamansir: May we calculate/fire this only if we have subscribers for this event?

src/anm/render.js
```diff
@@ -63,6 +68,13 @@ function r_loop(ctx, player, anim, before, after, before_render, after_render) {
         if (!after(time)) return;
     }
 
+    //increase the counter and fire the event if necessary
+    timeCounters[player.id] += dt;
+    if (timeCounters[player.id] >= TIME_UPDATE_TIME) {
```
shamansir: `TIME_UPDATE_DELTA`?)

shamansir:
In general, lgtm!

src/anm/interop/playerjs-io.js
mcm69: i'm afraid not, we're stuck with player.js spec

src/anm/render.js
mcm69: indeed, sounds better, my first implementation used a number of frames to fire the event, hence the strange name

src/anm/player.js
mcm69: maybe, but this method is used in `.volume()` anyway, so it would just complicate things

src/anm/interop/playerjs-io.js
shamansir: ok, np

src/anm/player.js
shamansir: In the best case I imagined `mute/unmute/volume` methods could be added from inside of `audio.js` to `Player.prototype` and `Controls.prototype`. Though nowadays audio is not so very separate from Player as before, and it could complicate things more than improve, so whatever :).

shamansir:
Is it ok to merge now? Can't see `player.seek` here :)

mcm69:
sorry, will be here soon ;)

mcm69:
should be good now

src/anm/render.js
mcm69: I'll make a note to revisit this later

shamansir:
Thank you!

PR #285, opened 2015-04-10
[PLAY-670] fix seeking via controls

mcm69:


shamansir:
Seems it could be merged only after #283 since contains commits from there.

mcm69:
hmm, i wonder how these commits got there, must have branched from a wrong source /shrug

mcm69:
let's cancel this one and do a clean PR without rogue commits

PR #288, opened 2015-04-17
Clean up Jakefile, use Closure Compiler instead of UglifyJS (and rewrite deployment tasks)

mcm69:
@shamansir plz review


Jakefile
```diff
@@ -1191,7 +850,7 @@ task('_build-file', { async: true }, function() {
       ].join(' ')
     ], EXEC_OPTS);
     _getCommintHash.on('stdout', function(COMMIT_INFO) {
-        var BUILD_TIME = _extended_build_time(),
+        var BUILD_TIME = _extended_build_time();
```
shamansir: Please either leave comma alone or add `var` to a next line :)

mcm69: `COMMIT_INFO` is already defined as the function parameter, thank god for the linter ;)

src/anm/utils.js
```diff
@@ -12,7 +12,7 @@ is.defined = function(v) {
 is.finite = global.isFinite;
 is.nan = global.isNaN;
 is.arr = Array.isArray;
-is.int = function(n) {
+is.integer = function(n) {
```
shamansir: All functions name inside `is` object consist of three-letter (okay, except `is.defined`, `is.finite` and `is.modifier` :) ), I think then it's better to rename them all to full names, for the sake of consistency.

Jakefile
shamansir: uh, my common mistake. then please remove an indentation here and move it a line above, so it will be visually in the other place for me, not to fix it _improperly_ :).

shamansir:
Looks cool, thank you!

shamansir:
@mcm69 Please also update instructions in `README.md` and wherever you find `push-version` with grep or whatever :)

Jakefile
mcm69: sure ;)

src/anm/utils.js
mcm69: I renamed `is.int` because closure compiler complained about usage of reserved word `int`, other methods are fine for now. I'll make a note to rename them later, consistency is nice, yeah.

PR #290, opened 2015-04-22
fix mishandling of keyboard events

mcm69:


mcm69:
@shamansir it previously threw errors when a keyboard event (with no `pos` obviously) arrived there. I've changed the function to filter just mouse events, let me know if we need the kb events here too.

shamansir:
@mcm69 We don't, seems previous variant of this code messed up my mind, thank you!

PR #293, opened 2015-04-29
[PLAY-681] Player doesn't show an object with a Stroke and without a Fill

mcm69:


src/import/animatron-importer.js
```diff
@@ -387,7 +389,7 @@ Import.leaf = function(type, src, parent, anim) {
     var trg = new Element();
     var hasUrl = !!src[1];
     if (!hasUrl &&
-        (type === TYPE_IMAGE || TYPE_AUDIO || TYPE_VIDEO)) {
+        (type === TYPE_IMAGE || type === TYPE_AUDIO || type === TYPE_VIDEO)) {
```
mcm69: ¯\\_(ツ)_/¯

skavish: wow

skavish:
so Path2D was not actually used, right? because I remember there was some speed improv (claimed)

mcm69:
I'm not sure when the typo in path2d.js was introduced - the test results are true (there were no usep2d check in the test player dist). So yes, it wasn't used for some time

PR #294, opened 2015-04-29
Revive sandbox + New API changes

shamansir:
`Tween` / `Modifier` API improved a lot, now they're proper classes instead of function wrappers, and `Tween` is actually just extends `Modifier`. Also, nice names for tweens and easings instead of weird constants.
Sandbox now uses new API, including these changes.
First examples were updated to a new API, next examples will be updated later.

Will add it to `gh-pages` branch few hours later.

### DIFF
--- a/src/anm/animation/tween.js
+++ b/src/anm/animation/tween.js
@@ -63,7 +63,7 @@ function Tween(tween_type, data) {
     } else {
         var tween_def = _Tweens[tween_type];
         tween_f = tween_def.func;
-        // we update `modifier.$tween` function each time when tween data was updated with `from`/`to`/`values` methods
+        // we update `modifier.$tween` function each time when tween data was changed with `from`/`to`/`values` methods
         mod_f = function(t, dt, duration) { if (this.$tween) this.$tween.call(this, t, dt, duration); };
         mod_f.tween = tween_type;
         from_f = tween_def.from;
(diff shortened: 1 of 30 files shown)


shamansir:
@mcm69 please review

shamansir:
I will merge it to show it in action, it is still a subject to review, will do a separate PR for fixes.

src/anm/animation/tween.js
```diff
@@ -27,152 +34,284 @@ var errors = require('../errors.js');
...
-               };
-    mod.data = data_block_fn; // FIXME
+var Tween = Modifier;
+
+function createTween(type, value) {
+    var mod = new Modifier(null, C.MOD_TWEEN); // mod.func will be set to `null` first
+    var def = Registry[type];
+    mod.def = def;
+    mod.func = function(t, dt, duration) { if (mod.$tween) mod.$tween.call(this, t, dt, duration); };
+    mod.is_tween = true;
```
mcm69: what's this for, isn't it already created as `MOD_TWEEN`?

src/anm/animation/tween.js
```diff
@@ -27,152 +34,284 @@ var errors = require('../errors.js');
...
+Tween.register = function(type, definition) {
+    definition = is.fun(definition) ? { func: definition } : definition;
+    definition.from = definition.from || Tween.DEFAULT_FROM;
+    definition.to   = definition.to   || Tween.DEFAULT_TO;
+    Registry[type] = definition;
+    Tween[type] = function(value) { return createTween(type, value); };
+}
 
-var Tweens = {};
+Tween._$ = createTween;
```
mcm69: what's this?

src/anm/animation/tween.js
shamansir: true, will remove it. thanks!

src/anm/animation/tween.js
shamansir: Probably it's not needed to be exposed. It was planned to act similar to `Element._$`, which allows user to do something like `var element = Element._$` and then do nice chaining like: `element(<name>).band().fill()...`. So user decides what name should the chain root have.

PR #298, opened 2015-05-05
Interactivity fixes, p.I

shamansir:


sandbox/interactivity-sandbox.js
```diff
@@ -161,10 +166,20 @@ function sandbox() {
...
 }
 
 function load_last_code() {
     if (!localStorage) throw new Error('Local storage support required');
-    return localStorage.getItem('_current_code') || '';
+    return localStorage.getItem('_i_current_code') || '';
+}
+
+function save_current_json(json) {
+    if (!localStorage) throw new Error('Local storage support required');
```
mcm69: http://caniuse.com/#search=localstorage
don't think that is necessary. also, in the event `localStorage` is not defined, you'll get a `ReferenceError` for an undefined element here, so `window.localStorage` will have to be used.

shamansir: true. removed it. thanks!

src/anm/animation/animation.js
```diff
@@ -356,15 +357,43 @@ Animation.prototype.unsubscribeEvents = function(canvas) {
 };
 
 Animation.prototype.handle__x = function(type, evt) {
+    var anim = this;
     if (events.mouse(type)) {
         var pos = evt.pos;
-        this.each(function(child) {
+        var foundTarget = false;
+        anim.each(function(child) {
```
shamansir: This algorithm confuses me with it's even though it checks bounds first and mouse events are passed only if `options.handleEvents` is `true` and if element is subscribed to them. @mcm69 Probably you have some thoughts on improvement?

sandbox/interactivity-sandbox.js
RobertoUa: also some people can disable localStorage in browser settings. Also there's could be problems in some browsers in incognito mode.

shamansir: It's just a page for internal testing, it's not a big deal! :)

mcm69: i see someone is bored in Asia ;)

skavish: @RobertoUa go back to the beach :)

PR #300, opened 2015-05-06
Check if element subscribed to mouseOut prior to firing an event

shamansir:


shamansir:
This commit was merged in `test` branch accidentally, so I will merge it alone.

PR #301, opened 2015-05-07
Improvements for Animation/Element.find and Animation/Element.findAll

shamansir:
Add the ability to search with path in a way like: `/root/sub-element/some-name`, or even with indexes (they depend on the order elements were added to a tree): `/:1/:2/some-name`.


shamansir:
Tests are improperly running at Travis, I temporary disabled it. I plan to fix them with next PRs.

PR #303, opened 2015-05-12
[play-639] postpone the load completion

mcm69:
When the resources are loaded from cache, the player’s `stop` method and some of the other updates are fired before the elements get to know they’re ready. This leads to all the kinds of weird things like the thumbnail being drawn only with vector shapes and not images, like in PLAY-639.

This can be remedied with a rewrite of the ResourceManager (which should be done later anyway), but a simple and working method to fix it right away is just let the event loop do its job before calling `stop` and considering the load complete. This is how it is done here.

### DIFF
--- a/src/anm/player.js
+++ b/src/anm/player.js
@@ -375,20 +375,22 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
                     if (player.anim === result) { // avoid race condition when there were two requests
                         // to load different animations and first one finished loading
                         // after the second one
-                        player.state.happens = C.LOADING;
-                        player.fire(C.S_CHANGE_STATE, C.LOADING);
-                        player.fire(C.S_LOAD, result);
-                        player._updateMediaVolumes();
-                        if (!player.handleEvents) player.stop();
-                        player._callPostpones();
-                        if (callback) callback.call(player, result);
-                        // player may appear already playing something if autoPlay or a similar time-jump
-                        // flag was set from some different source of options (async, for example),
-                        // then the rule (for the moment) is: last one wins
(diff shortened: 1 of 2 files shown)


src/anm/player.js
```diff
@@ -375,20 +375,22 @@ Player.prototype.load = function(arg1, arg2, arg3, arg4) {
...
-                        player._callPostpones();
-                        if (callback) callback.call(player, result);
-                        // player may appear already playing something if autoPlay or a similar time-jump
-                        // flag was set from some different source of options (async, for example),
-                        // then the rule (for the moment) is: last one wins
-                        if (player.autoPlay) {
-                            if (player.state.happens === C.PLAYING) player.stop();
-                            player.play();
-                        }
+                        utils.postpone(function(){
```
shamansir: Actually there are already some postponing mechanics which delay the calls to `play`/`pause`/`stop`/... methods util the resources will finish loading (no matter, from cache or not), i.e. [here](https://github.com/Animatron/player/blob/e0ed315637c317bdbf7493430eb0250ea3879a7c/src/anm/player.js#L569). Following the description, I suppose they are somehow failing in this case. I could be wrong, though—have you considered these methods to fail or it's the different problem?

mcm69: They're called synchronously, and we need an event loop pass to notify the elements they're ready. 



—

shamansir: Got it, thanks! I propose to rename `utils.propose` later then, since, as for me, it's actually not _postponing_ in this case, but things used nearby have the similar name. `yield`, may be?

shamansir: ...or `queue`...

PR #305, opened 2015-05-13
Auto-play if Animation.hasScripting is true

shamansir:


mcm69:
Can you explain me the reasoning behind this?

shamansir:
We do interactivity now. So projects with event handlers should have no controls and auto-play when published (in `publish.html` page) or in ether case controls will capture all the mouse events. `.hasScripting` (the previous name for _interactivity_ :) ) flag is set with Animatron Importer when project like this was received.

In the best case, `autoPlay` and `controlsEnabled` options should be set in `publish.html`/`.js` for these projects (and no weird assumptions should be done in the importer), so controls will be disabled from the start, before the loading process, in the proper way. But it has to be done at the server side, so we decided to leave this flag here temporary.

mcm69:
> So projects with event handlers should have no controls and auto-play when published
can't say i like this

shamansir:
Why not? :). If it's a banner or something interactive, it could be hard for user to click shapes inside. Or it's regarding auto-play? Auto-play is the result of disabled controls: user has no "Play" button, so unable to start "playing" the scene (which is infinite, by the way—it never stops).

mcm69:
yeah, sounds reasonable now, thanks for the explanation

PR #306, opened 2015-05-19
A number of fixes for mouse event handling.

shamansir:
Mouse event check now goes over the children in reverse order, so top elements (with higher _virtual_ z-index) are checked first. First element found for a `mouseclick` and matched by its time is the only one that handles this event. Other mouse events still check all matched elements. If element is not subscribed to the mouse event, it is passed to the first `parent` which is subscribed

Also includes:
* weird `handle__x` from `events.js` is now renamed to a friendly `filterEvent`;
* `Animation.each`, `Animation.traverse`, `Element.each`, `utils.keys` can now be exited by explicitly returning `false` from the iteration function;
* `Animation.reverseEach`, `Element.reverseEach`, for searches which should first visit top elements _(or should `.each` always work like that?)_;
* `element.inside()` improvements;
* `animation.time` now stores the last time it was rendered, or `null`;
* `element.cur_t` / `element.cur_rt` instance properties now store the local time of last render (considering all the jumps in time could happen), or `null`;
* fix for `player.drawAt` to check `player.infiniteDuration` flag before firing the error about duration (happened in sandbox when you move it to background tab and return later, so it un-pauses and starts playing from where it was stopped at, but checked if time matches animation duration)

### DIFF
--- a/src/anm/animation/animation.js
+++ b/src/anm/animation/animation.js
@@ -388,7 +388,7 @@ Animation.prototype.filterEvent = function(type, evt) {
         var targetFound = false;
         var moSubscriber = null; // mouse-out subscriber
         if (type === 'mouseclick') console.log(':::: start checking for click at ', pos.x, pos.y);
-        anim.each(function(child) {
+        anim.each(function(child) { // iterates over
             child.inside(pos, function(elm) { // filter elements
                 if (type === 'mouseclick') console.log('checking:', elm.name, ', parent:', elm.parent ? elm.parent.name : 'None');
                 if (type === 'mouseclick') {
@@ -402,13 +402,14 @@ Animation.prototype.filterEvent = function(type, evt) {
                 if (type !== 'mousemove') {
                     if (subscriber) subscriber.fire(type, evt);
                 } else { // type === 'mousemove'
+
                     // check mouseover/mouseout
                     if (!anim.__lastOverElm) {
                         // mouse moved over this element first time
                         anim.__lastOverElm = elm; // not a subscriber!
                         if (subscriber) {
                             subscriber.fire('mouseover', evt);
-                            subscriber.fire(type, evt); // fire mousemove next to mouseover
+                            subscriber.fire(type, evt); // fire this mousemove next to mouseover
                         }
                     } else {
                         if (elm.id === anim.__lastOverElm.id) { // mouse is still over this element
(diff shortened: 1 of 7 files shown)


shamansir:
@mcm69 @skavish @kirillik please review.

src/anm/animation/animation.js
```diff
@@ -166,11 +167,34 @@ Animation.prototype.traverse = function(visitor, data) {
  *
  * @param {Function} visitor
  * @param {anm.Element} visitor.child
+ * @param {Boolean} visitor.return if `false` returned, stops the iteration. no-`return` or empty `return` both considered `true`.
  * @param {Object} [data]
  */
 Animation.prototype.each = function(visitor, data) {
     for (var i = 0, tlen = this.tree.length; i < tlen; i++) {
-        visitor(this.tree[i], data);
+        if (visitor(this.tree[i], data) === false) break;
```
skavish: comparing to false is somewhat bad on many levels

skavish:
lgtm considering that I am not sure I understood all of that

src/anm/animation/animation.js
kirillk: Needless to say that typically visitors or any types of consumers return
true saying "it's done", like, "consumed"

shamansir: for JavaScript it's a bit different since if you omit `return` (since it's quite easy to forget it), it's commonly considered _truthy_, and `false` value may only be achieved with explicitly returning `false`. It's only regarding the return values from functions. At least it was for `jQuery.each` and it's there for different implementations.

Though, in modern Javascript there's no way to break out of `Array.each` (probably for the sake of clean code, indeed), but there's a special [`Array.every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) method which works differently.

@mcm69 what would you prefer in this case?

kirillk: What is easy to forget must be covered with tests. Period.

shamansir: Tests are not the universal pill. Especially in JS. Sometimes you just need to trust yourself.

Again, it's a matter of decision, not a test. Returning `false` to skip element is quite common practice for JS, also it's properly noticed in docs. And, it's common to omit `return` in dynamic languages. You can't test users' code, who usually omit/forget `return` (like in 90% cases), especially while using `.each`.

kirillk: You need to write test test that makes sure the element is found, this is not some abstract “testing users’ code”
As long as you have tests for this code I’m fine even with returning false, it’s less evil than having no tests *especially* in dynamic languages

shamansir: Sure, there are tests for that.

PR #307, opened 2015-05-21
[do not merge] testing modules separately

mcm69:
@shamansir @skavish take a look - this seems to be an easy and working way to test player modules without having to mock the entire `engine` or any other module.

### DIFF
--- a/spec/media/audio.spec.js
+++ b/spec/media/audio.spec.js
@@ -1,0 +1,46 @@
+
+var nop = function(){}, empty = {};
+
+var proxyquire = require('proxyquire').noCallThru();
+var stubs = {
+    '../conf.js': empty,
+    '../log.js': empty,
+    '../utils.js': empty,
+    '../constants.js': empty,
+    'engine': {
+        isLocal: false,
+        isHttps: false,
+        createAudio: function() {
+            return {};
+        },
+        unsubscribeElementEvents: nop,
+        subscribeElementEvents: nop,
+        createSource: function() {
+            return {};
+        },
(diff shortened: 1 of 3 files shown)


mcm69:
use `jake test` or `jasmine` to, well, test.

spec/media/audio.spec.js
```diff
@@ -0,0 +1,46 @@
...
+        },
+        unsubscribeElementEvents: nop,
+        subscribeElementEvents: nop,
+        createSource: function() {
+            return {};
+        },
+        appendToBody: nop
+    },
+    '../resource_manager.js': empty
+};
```
shamansir: Very nice idea in general, but I suppose for other modules, rather than audio, engine mocks could appear quite huge :).On the other hand, it's probably a way faster (and cleaner?) to start testing this way, plus in the end, we'll get several mocks in every spec-file we could blend into one in the end. I like the fact every spec is independent this way.

spec/search.spec.js
```diff
@@ -1,281 +1,281 @@
...
-        expect(subRootElement.findAll('/:5')[0].name).toEqual('index-5');
-
-        expect(subRootElement.find('/:7')).toBeNull();
-        expect(subRootElement.findAll('/:7')).toBeEmpty();
-
-
-    });
-
-});
+//if (typeof anm === 'undefined') {
```
shamansir: If this test will work, I'd prefer your way for now.

shamansir:
Seems to be obsolete due to successful Karma configuration

PR #308, opened 2015-05-25
[PLAY-665] fixed - change how injecting styles work

mcm69:
- The injected `style` and also the text measurer now have IDs for easier identification and debugging;
- Previously, when a player was detached, both the instance-specific and _generic_ CSS was removed, so once you detached a player, you lost the CSS configuration for the controls, which led to weird layout issues. This is no more;
- The player preferred to create a CSS rule for each an every canvas/wrapper and write the width/height and other data into the rule, rather than using elements' inline style. This was horribly confusing and a pain to debug, so it was removed.

@shamansir @skavish please review.

### DIFF
--- a/src/engine/dom-engine.js
+++ b/src/engine/dom-engine.js
@@ -263,7 +263,7 @@ $DE.ensureGlobalStylesInjected = function() {
     //if (!($doc.readyState === "complete")) return;
     var stylesTag = $doc.getElementById('anm-player-styles');
     if (!stylesTag) {
-        $DE.createStyle();
+        stylesTag = $DE.createStyle();
         stylesTag.id = 'anm-player-styles';
         // TODO: inject as first element?
         var head = $doc.getElementsByTagName("head")[0];
--- a/src/engine/dom-engine.js
+++ b/src/engine/dom-engine.js
@@ -261,16 +261,15 @@ $DE.styling = {
 $DE.ensureGlobalStylesInjected = function() {
     if ($DE.__stylesTag) return;
     //if (!($doc.readyState === "complete")) return;
-    var stylesTag = $DE.createStyle();
+    var stylesTag = $doc.getElementById('anm-player-styles');
+    if (!stylesTag) {
+        $DE.createStyle();
+        stylesTag.id = 'anm-player-styles';
+        // TODO: inject as first element?
+        var head = $doc.getElementsByTagName("head")[0];
+        head.appendChild(stylesTag);
 
-    // TODO: inject as first element?
-    var head = $doc.getElementsByTagName("head")[0];
-    if (!head) throw new Error('anm.Player requires <head> tag to exist in the document to inject CSS there');
-    head.appendChild(stylesTag);
-    // TODO: inject as first element?
-    // var head = $doc.getElementsByTagName("head")[0];
-    // head.insertBefore(stylesTag, head.firstChild);
-
+    }
     $DE.__stylesTag = stylesTag;
 };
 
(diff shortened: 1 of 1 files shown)


shamansir:
> The player preferred to create a CSS rule for each an every canvas/wrapper and write the width/height and other data into the rule, rather than using elements' inline style. 
It was done to give user the ability to redefine instance styles without using `!important` or javascript. I.e. redefine size of the specific player, when it's injected w/o any javascript (using HTML attributes).

src/engine/dom-engine.js
```diff
@@ -288,27 +285,21 @@ $DE.injectElementStyles = function(elm, general_class, instance_class) {
...
 };
 
-$DE.__textBuf = null;
+$DE.__textBuf = $doc.getElementById('anm-text-measurer');
 $DE.createTextMeasurer = function() {
     var buff = $DE.__textBuf;
     if (!buff) {
-        /* FIXME: dispose buffer when text is removed from animation */
         $DE.onDocReady(function(){
           var div = $doc.createElement('div');
```
shamansir: If `player.play` was called in somewhere in the document (or exactly after the `canvas` it uses, but before the `body` end), `buff` has a chance not to be ready for this time.

src/engine/dom-engine.js
```diff
@@ -288,27 +285,21 @@ $DE.injectElementStyles = function(elm, general_class, instance_class) {
...
-        instance_rule_idx = (styles.insertRule || styles.addRule).call(styles, '.' +instance_class + '{}', rules.length);
-    var elm_rules = [ rules[general_rule_idx],
-                      rules[instance_rule_idx] ];
-    props.gen_rule  = elm_rules[0];
-    props.inst_rule = elm_rules[1];
-    return elm_rules;
 };
 
-$DE.__textBuf = null;
+$DE.__textBuf = $doc.getElementById('anm-text-measurer');
```
shamansir: Why could it be already there when this script is executed (usually placed in `head`)?

src/engine/dom-engine.js
```diff
@@ -288,27 +285,21 @@ $DE.injectElementStyles = function(elm, general_class, instance_class) {
     var props = $DE.getAnmProps(elm);
     props.gen_class  = general_class;
     props.inst_class = instance_class;
```
shamansir: Seems this data is never used now, so `injectElementStyles` is not needed at all?

mcm69:
could you give me an example of how this was supposed to work before (i mean redefining the attributes?)

src/engine/dom-engine.js
mcm69: that's actually another bug and another story, i'll have that fixed next

src/engine/dom-engine.js
mcm69: i suppose it couldn't, it's here for the sake of uniformity and for an unlikely case there are different player versions on a same page (though i suspect we'll need more changes to actually support the latter).

src/engine/dom-engine.js
mcm69: that's indeed a leftover. thanks

shamansir:
If container for a player looks like: `<div id="some-id" anm-player-target="true" anm-src="..." ...></div>`

Then it gets the classes: `<div id="some-id" class="anm-player anm-player-some-id">`, so user may do this in CSS:

```css
.anm-player-some-id { /* or `#some-id` */
    width: ...;
    height: ...;
    border: ...;   
}
```

and override styles this way. At least it was planned to work like this :)

mcm69:
i see your point. i'm not sure it ever worked though - the user stylesheet had the same priority as player's and the latter was added to the end of the `head`.

for width/height there are still `anm-width` and `anm-height` and we don't exactly mess with other properties, so you can define the `border` just as you could before ;)

src/engine/dom-engine.js
shamansir: ok, got it. it may happen if player is included with several `iframe`s. thanks.

shamansir:
I'm not sure it ever worked though — me too, honestly, had no ability to test it hardly. Ok, I do agree, probably it has no sense :).

PR #310, opened 2015-06-02
A number of fixes to masks timing calculation.

shamansir:


shamansir:
NB: This PR is to merge into `master`, not a `develop`.

shamansir:
@skavish Is it ok to merge?

src/anm/animation/element.js
```diff
@@ -845,12 +845,15 @@ Element.prototype.render = function(ctx, gtime, dt) {
 
                 var mask = this.$mask;
 
+                var mask_ltime = mask.ltime(gtime),
+                    mask_gtime = mask.gtime(gtime);
```
shamansir: `mask_ltime` should be here instead of `gtime`, will fix that.

PR #316, opened 2015-06-16
Move player's CSS to a separate file

mcm69:
This makes the player’s CSS a lot more maintainable and visible. Will help with the end screen too, as it is a DOM object and requires a lot of styling.

One caveat: if you use `watchify` when developing, you should now launch it like this now:
`watchify -t browserify-css src/main.js -o dist/player.js` or else it will not build.


mcm69:
@shamansir please review

shamansir:
Looks cool! Thank you!

PR #319, opened 2015-06-19
[PLAY-636] add controlsInvisible option

mcm69:
With this option, one can hide the controls but still click to play/pause the animation.

@shamansir plz review

### DIFF
--- a/src/anm/ui/controls.js
+++ b/src/anm/ui/controls.js
@@ -234,7 +234,7 @@ Controls.prototype.react = function() {
 
     //handle clicks in the bottom area, where the playhead
     //and mute buttons reside
-    if (Controls.isInProgressArea(coords, w, h)) {
+    if (!this.invisible && Controls.isInProgressArea(coords, w, h)) {
         if (coords.x > btnWidth && coords.x < w-btnWidth) {
             time = utils.roundTo(p.state.duration*(coords.x-btnWidth)/(w-2*btnWidth), 1);
             if (time > p.anim.duration) {
--- a/src/anm/ui/controls.js
+++ b/src/anm/ui/controls.js
@@ -25,6 +25,7 @@ function Controls(player) {
     this.bounds = [];
     this.theme = null;
     this.info = null;
+    this.invisible = player.controlsInvisible;
 
     this.state = {
         happens: C.NOTHING,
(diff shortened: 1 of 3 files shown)


shamansir:
Honestly I don't think we need making controls invisible for that, I have an issue in YouTrack to handle click on canvas when controls are disabled and there's no interactivity, as a signal to start the scene. So clicking on canvas will start the movie — it's just one mouseclick handler instead of an additional canvas. It won't, probably, support pausing, though. @skavish what do you think?

mcm69:
I don't think handling events on canvas is a good idea, all responsibility regarding that is implemented in controls and should stay there.

shamansir:
Ok, let's see what Dima will say. 

My point also is that for API end-user the "invisible controls" option would look strange, since user doesn't know the controls canvas is placed over the whole player area, user [probably] thinks it's just a button in the center and a tiny line at the bottom, and they become invisible, which makes no sense.

Also, when controls are disabled (as there's such option), there's of course should be a way to start a scene if auto-play is also off. It's just a call to `play()`/`stop()`, `Controls` class is mostly responsible for visualizing progress and buttons which, in their turn, control playback.

src/anm/ui/controls.js
```diff
@@ -154,6 +155,10 @@ Controls.prototype.render = function(gtime) {
         coords = state.mpos,
         time = state.time = player.state.time;
 
+    if (this.invisible) {
```
shamansir: I think it should be above both `this.rendering = true;` and `this.checkFade(dt)`.

skavish:
frankly I don't understand the argument. I checked the code and I don't see any extra canvas. what are we talking about here?

mcm69:
Anton is thinking of _player_ canvas handling the clicks when controls are disabled/hidden, I think this should not be its responsibility.

shamansir:
That's not actually _extra_ canvas, it's just not disabled and still added to DOM when it's not shown. I suppose when user _disables) controls, and auto-play is off (by default), she still expects she'll be able to run it by clicking on the player canvas. I think there's no need in special option for that and no need to load controls only for that purpose.

mcm69:
the user does not give a damn how many elements the player creates nor it should be his business

shamansir:
(but Yura disabled the rendering if "invisible" flag is true, it worths mentioning)

shamansir:
This flag is a part of an API, it could be just the naming problem, as an API-user I don't actually get what "invisible controls" means.

shamansir:
Okay, I'll merge it for the sake of release)

PR #322, opened 2015-07-07
Fix orient to path (again)

shamansir:
Now parameter `t` is saved in Translate tween, so it's easy for Orient-to-path tween to have any band it likes, and still be correct. Why haven't I thought about it before?

### DIFF
--- a/src/anm/animation/tween.js
+++ b/src/anm/animation/tween.js
@@ -259,7 +259,7 @@ Tween.register(C.T_ROT_TO_PATH, {
             var path = this.$mpath;
             // when t equals exact 0, it is replaced with 0.001
             // or else returned angle would be 0
-            if (path) this.angle += path.tangentAt(t || 0.001);
+            if (path) this.angle = path.tangentAt(t || 0.001);
         }
     },
     from: nop, to: nop
--- a/src/anm/animation/tween.js
+++ b/src/anm/animation/tween.js
@@ -265,6 +265,17 @@ Tween.register(C.T_ROT_TO_PATH, {
     from: nop, to: nop
 });
 
+Tween.register(C.T_MRG_ROT_TO_PATH, function(values) {
+    var _from = values[0],
+        to = values[1];
+    return function(t) {
+        var path = this.$mpath;
+        var angle = _from * (1.0 - t) + to * t;
+        if (path) angle += path.tangentAt(t || 0.001);
+        this.angle = angle;
+    }
+});
+
 Tween.register(C.T_ALPHA, function(values) {
     var _from = values[0],
         to = values[1];
--- a/src/anm/animation/tween.js
+++ b/src/anm/animation/tween.js
@@ -259,21 +259,10 @@ Tween.register(C.T_ROT_TO_PATH, {
             var path = this.$mpath;
             // when t equals exact 0, it is replaced with 0.001
             // or else returned angle would be 0
-            if (path) this.angle = path.tangentAt(t || 0.001);
+            if (path) this.angle += path.tangentAt(t || 0.001);
         }
     },
     from: nop, to: nop
-});
-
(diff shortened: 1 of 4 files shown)


shamansir:
@skavish please take a look.

src/anm/animation/tween.js
```diff
@@ -259,7 +260,7 @@ Tween.register(C.T_ROT_TO_PATH, {
             var path = this.$mpath;
             // when t equals exact 0, it is replaced with 0.001
             // or else returned angle would be 0
-            if (path) this.angle += path.tangentAt(t || 0.001);
+            if (path) this.angle += path.tangentAt(this.$mpath_t || 0.001);
```
skavish: why 0.001?

shamansir: should it be less? for a value of exact `0`, tangent is always `0`

src/import/animatron-importer.js
```diff
@@ -284,27 +284,34 @@ Import.branch = function(type, src, all, anim) {
...
             }
-            if (translates && (flags & L_ROT_TO_PATH)) {
-                var rtp_tween;
-                for (ti = 0, til = translates.length; ti < til; ti++) {
-                    rtp_tween = Tween[C.T_ROT_TO_PATH]();
-                    if (translates[ti].$band) rtp_tween.band(translates[ti].$band);
-                    if (translates[ti].$easing) rtp_tween.easing(translates[ti].$easing);
-                    ltrg.tween(rtp_tween);
+            if ((flags & L_ROT_TO_PATH) && rotate_tweens) {
+                if ((first_rotate > 0) && (first_rotate < Infinity)) {
```
skavish: > ? or may be >= ?

src/import/animatron-importer.js
```diff
@@ -284,27 +284,34 @@ Import.branch = function(type, src, all, anim) {
...
+                    ltrg.tween(Tween.rotate().start(0).stop(first_rotate)
+                                             .from(0).to(0));
+                }
+                if ((last_rotate > 0) && (last_rotate < Infinity)) {
```
skavish: same here

src/import/animatron-importer.js
shamansir: It has no sense to add a tween which starts and _stops_ both at 0 (see line below)

src/import/animatron-importer.js
```diff
@@ -284,27 +284,34 @@ Import.branch = function(type, src, all, anim) {
...
+                    ltrg.tween(Tween.rotate().start(last_rotate).stop(ltrg.lband[1] - ltrg.lband[0])
```
shamansir: ...or starts at `Infinity` in this case

src/anm/animation/tween.js
skavish: why calculate it at all?
```
....  = this.$mpath_t? path.tangentAt(this.$mpath_t): 0;
```

shamansir: It needs to be something _other than 0_ in case if `mpath_t` is 0, so i.e. in very first frame or the animation, the shape will be properly oriented to path (there were several issues on when player shows first frame wrong in "thumbnail" (at time `0`) because shape is rotated to 0 degrees while it shouldn't).

shamansir: I remember you suggested that to me as a fix for it :)

skavish: well, i guess, it makes sense in this case. in the editor we fixed in path.tangentAt, it skips all static parts

shamansir: Honestly, it's also not about static parts :) static parts are skipped [here](https://github.com/Animatron/player/pull/322#discussion-diff-34006892R221) instead. It's just because tangent at the very start point of any path is always 0.

PR #325, opened 2015-07-22
fix Back In, Back Out, Instant In, Instant Out easings, introduce bounded easings

shamansir:


shamansir:
@alexeypegov notice that one please, I've seen you copied easings from here, but they weren't totally correct.

skavish:
@alexeypegov you should have copied easings from editor

alexeypegov:
@skavish well, I copied them from player

PR #327, opened 2015-09-03
WIP. Player analytics

RobertoUa:
do not merge yet, WIP

### DIFF
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -8,7 +8,7 @@ var Analytics = function () {
         beacon = null,
         animatronUrl = utils.makeApiUrl('/analytics/player');
 
-    self.enabled = animatronUrl == null || animatronUrl.indexOf('animatron-test') >= 0;
+    self.enabled = animatronUrl != null || animatronUrl.indexOf('animatron-test') >= 0;
     self.queue = [];
 
     var event = function () {
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -6,7 +6,7 @@ var Analytics = function () {
         supportSendBeacon = !!navigator.sendBeacon,
         timeout = supportSendBeacon ? 2000 : 1000,
         beacon = null,
-        animatronUrl = utils.makeApiUrl('/analytics');
+        animatronUrl = utils.makeApiUrl('/analytics/player');
 
     self.queue = [];
 
@@ -19,7 +19,7 @@ var Analytics = function () {
                 navigator.sendBeacon(animatronUrl, data);
                 setTimeout(event, timeout);
             } else {
-                var trackUrl = animatronUrl + '?player=' + encodeURIComponent(data);
+                var trackUrl = animatronUrl + '?data=' + encodeURIComponent(data);
                 sendViaGif(trackUrl);
             }
         } else {
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -35,7 +35,7 @@ var Analytics = function () {
     };
     if (self.enabled) {
         if (!supportSendBeacon) {
-            event();
+            engine.onDocReady(event);
         }
         window.addEventListener('unload', event, false);
     }
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -1,0 +1,60 @@
+var engine = require('engine');
+
+var Analytics = function () {
+    var self = this,
+        timeout = 1000,
+        beacon = null,
+        animatronUrl = 'http://localhost:8080/analytics?';
+
+    self.queue = [];
+
+    var event = function () {
+        if (self.queue.length > 0) {
+            var trackUrl = animatronUrl + 'player=' + encodeURIComponent(JSON.stringify(self.queue));
+            self.queue = [];
+
+            if (!beacon) {
+                beacon = engine.createStatImg();
+            }
+            beacon.src = trackUrl;
+            beacon.onerror = beacon.onload = function (e) {
+                console.log('here we are', trackUrl, e);
+                beacon.onerror = beacon.onload = null;
+                setTimeout(event, timeout);
+            }
+        } else {
+            setTimeout(event, timeout);
+        }
+    };
+    event();
+    window.addEventListener('beforeunload', event, false);
+
+    this.trackPlayingStart = this.trackPlayer('playing_start');
+    this.trackPlayingPause = this.trackPlayer('playing_pause');
+    this.trackPlayingComplete = this.trackPlayer('playing_complete');
+};
+
+Analytics.prototype.track = function track(name, opts) {
+    opts = opts || {};
+    opts.name = name;
+    opts.referer = document.referrer;
+    opts.lang = navigator.language || navigator.userLanguage;
+    this.queue.push(opts);
+};
+
+Analytics.prototype.trackPlayer = function trackPlayer(name) {
+    return function (player) {
+        var opts = {viewId: player.viewId, time: player.state.time};
+        this.track(name, opts);
+    }.bind(this);
+};
+
+
+Analytics.prototype.getObjectId = function () {
+    var timestamp = (new Date().getTime() / 1000 | 0).toString(16);
+    return timestamp + 'xxxxxxxxxxxxxxxx'.replace(/[x]/g, function () {
+            return (Math.random() * 16 | 0).toString(16);
+        }).toLowerCase();
+};
+
+module.exports = new Analytics();
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -8,7 +8,7 @@ var Analytics = function () {
         beacon = null,
         animatronUrl = utils.makeApiUrl('/analytics/player');
 
-    self.enabled = animatronUrl != null || animatronUrl.indexOf('animatron-test') >= 0;
+    self.enabled = animatronUrl != null && animatronUrl.indexOf('animatron-test') >= 0;
     self.queue = [];
 
     var event = function () {
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -8,6 +8,7 @@ var Analytics = function () {
         beacon = null,
         animatronUrl = utils.makeApiUrl('/analytics/player');
 
+    self.enabled = animatronUrl == null || animatronUrl.indexOf('animatron-test') >= 0;
     self.queue = [];
 
     var event = function () {
@@ -37,8 +38,10 @@ var Analytics = function () {
             setTimeout(event, timeout);
         }
     };
-    event();
-    window.addEventListener('unload', event, false);
+    if (self.enabled) {
+        event();
+        window.addEventListener('unload', event, false);
+    }
 
     this.trackPlayingStart = this.trackPlayer('playing_start');
     this.trackPlayingPause = this.trackPlayer('playing_pause');
@@ -56,7 +59,9 @@ Analytics.prototype.track = function track(name, opts) {
     opts.windowHeight = window.innerHeight;
     opts.windowWidth = window.innerWidth;
     opts.timestamp = new Date().getTime();
-    this.queue.push(opts);
+    if (this.enabled) {
+        this.queue.push(opts);
+    }
 };
 
 Analytics.prototype.trackPlayer = function trackPlayer(name) {
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -30,7 +30,7 @@ var Analytics = function () {
             beacon = engine.createStatImg();
         }
         beacon.src = trackUrl;
-        beacon.onerror = beacon.onload = function (e) {
+        beacon.onerror = beacon.onload = function () {
             beacon.onerror = beacon.onload = null;
             setTimeout(event, timeout);
         }
@@ -53,12 +53,13 @@ Analytics.prototype.track = function track(name, opts) {
     opts.screenWidth = screen.width;
     opts.windowHeight = window.innerHeight;
     opts.windowWidth = window.innerWidth;
+    opts.timestamp = new Date().getTime();
     this.queue.push(opts);
 };
 
 Analytics.prototype.trackPlayer = function trackPlayer(name) {
     return function (player) {
-        var opts = {viewId: player.viewId, time: player.state.time};
+        var opts = {viewId: player.viewId, projectId: player.anim.meta._anm_id, time: player.state.time};
         this.track(name, opts);
     }.bind(this);
 };
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -4,7 +4,7 @@ var engine = require('engine'),
 var Analytics = function () {
     var self = this,
         supportSendBeacon = !!navigator.sendBeacon,
-        timeout = supportSendBeacon ? 2000 : 1000,
+        timeout = 1000,
         beacon = null,
         animatronUrl = utils.makeApiUrl('/analytics/player');
 
@@ -18,28 +18,25 @@ var Analytics = function () {
 
             if (supportSendBeacon) {
                 navigator.sendBeacon(animatronUrl, data);
-                setTimeout(event, timeout);
             } else {
-                var trackUrl = animatronUrl + '?data=' + encodeURIComponent(data);
-                sendViaGif(trackUrl);
+                sendViaGif(data);
             }
-        } else {
+        }
+        if (!supportSendBeacon) {
             setTimeout(event, timeout);
         }
     };
 
-    var sendViaGif = function (trackUrl) {
+    var sendViaGif = function (data) {
         if (!beacon) {
             beacon = engine.createStatImg();
         }
-        beacon.src = trackUrl;
-        beacon.onerror = beacon.onload = function () {
-            beacon.onerror = beacon.onload = null;
-            setTimeout(event, timeout);
-        }
+        beacon.src = animatronUrl + '?data=' + encodeURIComponent(data);
     };
     if (self.enabled) {
-        event();
+        if (!supportSendBeacon) {
+            event();
+        }
         window.addEventListener('unload', event, false);
     }
 
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -5,19 +5,20 @@ var Analytics = function () {
         supportSendBeacon = !!navigator.sendBeacon,
         timeout = supportSendBeacon ? 2000 : 1000,
         beacon = null,
-        animatronUrl = 'http://localhost:8080/analytics?';
+        animatronUrl = 'http://localhost:8080/analytics';
 
     self.queue = [];
 
     var event = function () {
         if (self.queue.length > 0) {
-            var trackUrl = animatronUrl + 'player=' + encodeURIComponent(JSON.stringify(self.queue));
+            var data = JSON.stringify(self.queue);
             self.queue = [];
 
             if (supportSendBeacon) {
-                navigator.sendBeacon(trackUrl);
+                navigator.sendBeacon(animatronUrl, data);
                 setTimeout(event, timeout);
             } else {
+                var trackUrl = animatronUrl + '?player=' + encodeURIComponent(data);
                 sendViaGif(trackUrl);
             }
         } else {
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -2,7 +2,8 @@ var engine = require('engine');
 
 var Analytics = function () {
     var self = this,
-        timeout = 1000,
+        supportSendBeacon = !!navigator.sendBeacon,
+        timeout = supportSendBeacon ? 2000 : 1000,
         beacon = null,
         animatronUrl = 'http://localhost:8080/analytics?';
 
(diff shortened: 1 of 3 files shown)


shamansir:
What to do you think about [using ReqestIdleCallback for sending analytics data](https://developers.google.com/web/updates/2015/08/27/using-requestidlecallback#using-requestidlecallback-for-sending-analytics-data) ? (/cc @skavish)

RobertoUa:
@shamansir which browsers support it except for nightly chrome?

RobertoUa:
I'm using navigator.sendBeacon it's supported in chrome and firefox. And can be used when tab is closing.

shamansir:
We may try `requestIdleFallback` first, as a future possible web standard, and fallback to sendBeacon then if there's no support for it (and to immediate sending if there's no sendBeacon). But yes, it's too fresh on the other hand, spec is pushed 5 days ago, and it's a draft.

RobertoUa:
I don't think there will be a lot of benefit from it.

skavish:
lgtm

src/anm/analytics.js
```diff
@@ -0,0 +1,76 @@
+var engine = require('engine');
+
+var Analytics = function () {
+    var self = this,
+        supportSendBeacon = !!navigator.sendBeacon,
+        timeout = supportSendBeacon ? 2000 : 1000,
+        beacon = null,
+        animatronUrl = 'http://localhost:8080/analytics';
```
skavish: I don't see where it is configured for test and prod?

src/anm/analytics.js
```diff
@@ -0,0 +1,76 @@
+var engine = require('engine');
+
+var Analytics = function () {
+    var self = this,
+        supportSendBeacon = !!navigator.sendBeacon,
+        timeout = supportSendBeacon ? 2000 : 1000,
```
skavish: why timeouts are different?

RobertoUa: I would like not to use timeout at all for sendBeacon, and invoke the function only when tab is closing, but I'm afraid that it will not always work.

src/anm/analytics.js
RobertoUa: it's not yet. But if it's lgtm, i will add it and this can be merged.

RobertoUa:
@shamansir pls review my latest commit (d6de3f7). There's code I so did not want to touch :(

RobertoUa:
@skavish I wonder which additional data I should save. Like options passed to player (repeat, autostart, width&height, etc.)

src/anm/analytics.js
shamansir: As I get it from description, the point of `sendBeacon` is that it is guaranteed to be called. So I think it's safe not to use timeouts in case of `sendBeacon`.

RobertoUa: It's safe in some cases. But what if browser crashes or user goes offline or he just didn't close the tab(and then went offline or smth).

shamansir: >  The browser queues the request, and sends it “at the earliest available opportunity, but MAY prioritize the transmission of data lower compared to other network traffic.” (as per the W3C spec). sendBeacon returns `true` if the data was successfully submitted to the server, or `false` otherwise.
So you may use return value if it failed.

src/anm/analytics.js
```diff
@@ -0,0 +1,77 @@
...
+    self.queue = [];
+
+    var event = function () {
+        if (self.queue.length > 0) {
+            var data = JSON.stringify(self.queue);
+            self.queue = [];
+
+            if (supportSendBeacon) {
+                navigator.sendBeacon(animatronUrl, data);
+                setTimeout(event, timeout);
```
shamansir: I got it, you are not postponing the request itself here, but postpone the check. But if `sendBeacon` will work, but much later than your timeout (i.e. when user closed tab), which will also work, the things will get messy.

shamansir: (Yes, `queue` will be empty for `sendBeacon` call, but what's the point?)

src/anm/player.js
```diff
@@ -644,6 +648,8 @@ Player.prototype.pause = function() {
     player.fire(C.S_CHANGE_STATE, C.PAUSED);
     player.fire(C.S_PAUSE, state.time);
 
+    analytics.trackPlayingPause(player);
```
shamansir: I see no `trackPlayingStart `/`trackPlayingPause` in `analytics.js` by the way

shamansir: Uh, sorry, found them: https://github.com/Animatron/player/pull/327/files#diff-7f5481547366f17ae675b154345e696eR43

RobertoUa: :)

src/anm/analytics.js
RobertoUa: I know, but what I don't know what I can possibly use it for

src/anm/analytics.js
RobertoUa: sorry I didn't get your comments.

I'm not postponing, I'm scheduling the check and if queue is not empty do the request, preferably using `sendBeacon`

src/anm/analytics.js
shamansir: Actually, the more I see information about it, the more it also looks very new and unestablished (not as new as idleCallback, but still :) ).

src/anm/analytics.js
```diff
@@ -0,0 +1,77 @@
...
+        if (!beacon) {
+            beacon = engine.createStatImg();
+        }
+        beacon.src = trackUrl;
+        beacon.onerror = beacon.onload = function () {
+            beacon.onerror = beacon.onload = null;
+            setTimeout(event, timeout);
+        }
+    };
+    event();
```
shamansir: Why we call it here? "Just in case"? :)

RobertoUa: to start everything

src/anm/analytics.js
shamansir: Yes, I found that it subscribes to `unload` event below, sorry, got the point. Try to send, and then check if we sent it, okay :).

src/anm/analytics.js
RobertoUa: well, yeah. But it is supported by blink based browsers and firefox. And api compatibly are unlikely to broken, since it only sends POST request, what breaking change can be there?

src/anm/analytics.js
```diff
@@ -0,0 +1,77 @@
...
+    opts.name = name;
+    opts.referer = document.referrer;
+    opts.lang = navigator.language || navigator.userLanguage;
+    opts.url = location.href;
+    opts.screenHeight = screen.height;
+    opts.screenWidth = screen.width;
+    opts.windowHeight = window.innerHeight;
+    opts.windowWidth = window.innerWidth;
+    opts.timestamp = new Date().getTime();
+    this.queue.push(opts);
```
shamansir: Are we collecting this data (lang, url, screen) for every event happened?

RobertoUa: yes, why not? Denormalized data is much faster and easier to manipulate.

RobertoUa: Or you propose to aggregate it here for the request and flatten on the backend?

src/anm/analytics.js
shamansir: In this case `self.queue` will be empty, so with `else` clause in `event` function, you'll set the timeout. May be some events will be collected to this point, so it will probably send something (why you set the timeout if queue is empty, by the way?). Then, you subscribe to `unload` event, it will call the `event` function on unload with the remaining event which were yet not sent. 

This will work (some events will probably, if happened, be sent once after a timeout, and all the remaining ones on `unload`), yes, but is it really was intended to work this way?

src/anm/analytics.js
shamansir: Don't know, may be it's easier this way, yes. Just wanted to clarify.

shamansir:
Considering everything noted, is it OK to merge? :). I may merge it, so we test it in action (at test).

By the way, what do you think, is it possible to test it with Jasmine (oh no, what am I saying!) with simulating `onload` event? Since we hope to get real statistics. 

Also, it could have sense to give a client some UID, or make up a session ID, which will be included in every package, something to ensure every package (even with the same timestamp) is unique.

RobertoUa:
do not merge it until API changes are deployed to test https://github.com/Animatron/animatron-editor/pull/4568

I don't see what to test here, it's pretty straightforward. Probably only mock analytics and check that proper methods are called on play,pause,etc.

Server side changes on the other hand need tests.

shamansir:
> I don't see what to test here, it's pretty straightforward.

Please explain this then: https://github.com/Animatron/player/pull/327#discussion_r39033827. Everything with async calls, timeouts and mutable external state involved can't be "straightforward", as for me. And if we want real analytics, it's required for us to be sure that event sent twice won't be treated as two different events.

> Probably only mock analytics and check that proper methods are called on play,pause,etc.

May be also this, yes. For combinations of `play`/`pause`/etc.

shamansir:
Tests are broken: https://travis-ci.org/Animatron/player/#L325. Please fix.

Also, you haven't answered here: https://github.com/Animatron/player/pull/327#discussion_r39033827

RobertoUa:
omg, they were ok but this commit broke them. Something wrong either with tests or with CI. https://github.com/RobertoUa/player/commit/a886bc4ec0f8a242d083dc6e41fa3cc1562338de

src/anm/analytics.js
RobertoUa: I again did not get your question.
This is basically works like this:
event() starts event loop. Without it nothing will be sent to the server. Only on unload event and if sendBeacon is supported. Then if queue is empty => do nothing, schedule check for queue again, in one sec. If queue is non empty, send analytic data to the sever and again, schedule check for queue in one sec.

RobertoUa:
updated

src/anm/analytics.js
shamansir: If I simplify your code, we'll get something like this:

```javascript
function Analytics() {
    var queue = [];
    function event() {
        if (!queue.empty()) {
            sendData();
            queue = [];
        } else {
            setTimeout(event, 1000);
        }
   }
   event();
   window.on('unload', event);
}
```

So, when you call `new Analytics()`, probably even before all scripts loaded into a page, you'll also manually call `event()` from inside. Since nothing yet happened to send, you just created the instance, and queue _for sure_ will be empty. So, it will set for 1sec or 2sec timeout to try to send data again (and, there's a high possibility that nothing meaningful will happen again for this second, page just loads).

So, the question is "Why?" :).

RobertoUa: >So, the question is "Why?" :).

Because it doesn't matter. It's obvious and I thought about it. There would be a reason to complicate the code even more if check for queue.empty() would be super expensive. You know, "premature optimization is the root of all evil".

shamansir: It's "premature optimization" in your case, since it will never send any data, why call it? I'm just against of this particular call, since it does nothing useful at all. No need to "complicate code".

RobertoUa: what do you propose? Call it with timeout of 5 secs, or what?

src/anm/analytics.js
```diff
@@ -0,0 +1,82 @@
...
+
+            if (supportSendBeacon) {
+                navigator.sendBeacon(animatronUrl, data);
+                setTimeout(event, timeout);
+            } else {
+                var trackUrl = animatronUrl + '?data=' + encodeURIComponent(data);
+                sendViaGif(trackUrl);
+            }
+        } else {
+            setTimeout(event, timeout);
```
shamansir: And in this case `timeout` will be 2000 if `supportSendBeacon` is `true`. It's not even related here.

src/anm/analytics.js
shamansir: Don't call it at all, why? Call it just on `window.unload`, that's enough.

src/anm/analytics.js
RobertoUa: Why is it not related here?? Read code carefully pls.
It doesn't matter whether it is send by sendBeacon or with gif.

src/anm/analytics.js
RobertoUa: We already discussed it! It's not safe to call only on 'unload'!

src/anm/analytics.js
shamansir: First, why the timeout is ever different for `sendBeacon`? Just in case?

This call is performed when queue is empty, no matter was `sendBeacon` supported or not, why should timeout be different here?

src/anm/analytics.js
shamansir: So "to be safe" we call it before page (and any Player instance) loaded and send nothing anyway? Twice, due to timeout? It's not safe, it's meaningless, as for me.

src/anm/analytics.js
RobertoUa: I already answered that question. SendBeacon is supposed to work on unload, but it's not safe to only call it on the event. So I still use dirty checking but with higher timeout.

src/anm/analytics.js
shamansir: It could send data with large intervals (_not_ timeouts) in the background while page is running, if you want to be "safe". And send the remains on `window.unload`. But this will affect page performance. I also don't know why you think `window.unload` could not be called. If it's a crash, I think we don't care, 0,01% of statistics will be lost, I think it's ok (@skavish is it?). It's more important for statistics to be reliable.

RobertoUa: >  send nothing anyway? 

Nothing is send.

>Twice, due to timeout

What twice? Pls point me what is sent twice.

> meaningless

What meaningless? To start the event loop right away? I can do it on `dom ready` event, but what the benefit? 100x performance boost? It's just a check for an array length.

shamansir:
Ok, considering Volodya knows about https://github.com/Animatron/player/pull/327#discussion_r39161783 and I tried to describe my point in details, this PR could be merged if @skavish is also OK with that.

src/anm/analytics.js
RobertoUa: Ok, if you are sure that we could go ahead without any timeout for `sendBeacon`, then ok, i will do this.

src/anm/analytics.js
```diff
@@ -0,0 +1,82 @@
...
+
+
+Analytics.prototype.getObjectId = function () {
+    var timestamp = (new Date().getTime() / 1000 | 0).toString(16);
+    return timestamp + 'xxxxxxxxxxxxxxxx'.replace(/[x]/g, function () {
+            return (Math.random() * 16 | 0).toString(16);
+        }).toLowerCase();
+};
+
+module.exports = new Analytics();
```
shamansir: Here you call this `event()` function first time. And postpone it for one second.

RobertoUa: yes, I call it here first time and schedule the call again in one sec.

src/anm/analytics.js
```diff
@@ -0,0 +1,79 @@
...
+
+    var sendViaGif = function (data) {
+        if (!beacon) {
+            beacon = engine.createStatImg();
+        }
+        beacon.src = animatronUrl + '?data=' + encodeURIComponent(data);
+    };
+    if (self.enabled) {
+        if (!supportSendBeacon) {
+            engine.onDocReady(event);
```
RobertoUa: @shamansir is it ok? I see that it will be called if document is already ready. So it's easy enough to add.

RobertoUa:
@shamansir updated, review again :)

src/anm/analytics.js
shamansir: Sorry, but no :). _Which_ statistics exactly you want to send when document is ready? Or, ok, one second later. No player yet created. No scene yet loaded. And so no play or pause button yet pressed, no interactivity script performed. At all. For sure.

And there is no "initialization" code in `event` function, it only contains sending data or postponing of sending it for one or two damn seconds. I _may_ of course merge it, I just feel that for the sake of "safety" you do exactly what you call "premature optimization". 

I'd expect `event` call will be performed one single time on `window.unload`. No need in emptying the queue, just fill this queue with everything happened and send as one package. Or, use intervals instead. Or, for example, call it on `player.detach` for every player instance created. And I could be wrong, of course, but that's 30 lines of code and I still see no point in that. Sorry :(.

shamansir:
I did all I can. I am merging it, since it doesn't hardly affect player performance, at least I suppose. It's just the matter of logic. But I expect Dima @skavish to review it also, or we discuss it together, and you'll send improvements on this later. May be it will feel better and easier for you if it will be merged, at least I often feel. And 40 comments for one tiny PR, that's enough.

src/anm/analytics.js
RobertoUa: Omg, Anton. Like 10 times I already told how it works.
'Safety' was about running the event loop for sendBeacon. I removed it.

Now it is performed only one time if browser supports sendBeacon. But not all of them support it. So I need to send them every N seconds in such case.


And now, pls tell me what are benefits of starting event loop when player is loaded? That is what I call premature optimizations. Not checking for length of an array. It just saves nothing. Absolutely nothing. But it does indeed complicates code. For nothing. Can you understand my point now? Because you either don't get what I am saying or understand it differently. I have to tell the same things many times, again and again.

shamansir: Vova, could it be that you misused `setTimeout` instead of `setInterval`? If you'd use `setInterval` instead, things'd start to get a little more sense. 

The problem is not in checking if array is empty once more. And it's not an event loop. Once again.
An example with browser without `sendBeacon`.

1. Browser or tab opened by user with this page. 
1. Browser loads scripts.
1. `Analytics` object created. Queue is empty. Postpone `event` for one second. (Imagine one second here).
1. Somewhere here browser starts to build DOM.
1. One or two second pass.
1. `event` function is called thanks to timeout. Queue is empty.
1. DOM finished construction.
1. ...
1. Some time passes, Players created.
1. ...
1. A lot time passes, during that time user does actions, fills `queue` with events, now it's not empty.
1. ...
1. User decides to close tab or a browser or "crash" happens.
1. Probably (you say, not guaranteed) `window.unload` handler is called. If it did, all the collected events flush to the server.
1. Fin.

How the call at time `6.` affects _safety_? If `window.unload` call is not guaranteed, in case of failure, nothing will be sent anyway. Timeouts complicate code.

shamansir: `event` could be called just once with the same effect. That's what I am trying to explain. No async, no timeouts needed. Or, if you don't want it so, use `setInterval`.

RobertoUa: you could've just said to use setInterval and that's all. Previously I had to use setTimeout. But thanks to the recent changes I indeed can use setInterval.

src/anm/utils.js
```diff
@@ -252,6 +252,31 @@ function postpone(fn) {
     setTimeout(fn, 0);
 }
 
+function makeApiUrl(path, loadSrc) {
+    var prodHost = 'animatron.com',
+        testHost = 'animatron-test.com',
```
skavish: i don't like animatron-test appearing in the source code of open source player

RobertoUa: I don't like it very much too. I only copied the code, and didn't want to touch it very much, so I won't break anything.

src/anm/analytics.js
shamansir: The problem with `setInterval`, as I said, is that it could affect page performance, if `sendBeacon` is not supported. Especially canvas, especially if it will run once every second. Every five minutes + on `unload` — ok, probably.

Look at every `sendBeacon` example, they just [sum up and] push data _once_ in a large array and send it on `window.unload`. Unload will always perform. Just, if `sendBeacon` is not supported, sending data could "delay the unload or affect the the performance of the next navigation".

RobertoUa: Omg, Anton. I'm already using sendBeacon+unload.

>Look at every sendBeacon example, they just [sum up and] push data once in a large array and send it once on window.unload

Omg, that's exactly what I do! Read the code.

If you want me to use `sendBeacon` polyfill instead of `setInterval` JUST SAY THAT.

I will use XHR polyfill. And delete `setInterval` whatsoever. Do we have something that supports CORS in IE9, or do I have to do it myself?

RobertoUa: Sorry for being so mad. It's just so much discussion and words for such small task. Just call me next time ;)

shamansir: Every time I expect my comment will be the last.

```javascript
var events = {}; // could be an array, if you want

function trackEvent(name, stats) {
    if (!events[name]) events[name] = [];
    events[name].push(stats);
}

function getData() {
    return { UID: 'aaa0-bbbx-...', // there's utils.guid()
                 navigator: ...,  events: events };
}

window.addEventListener('unload', function(event) {
  var xhr,
      data = getData();

  if (navigator.sendBeacon) {
    navigator.sendBeacon('api/stats/report', data);
  } else {
    xhr = new XMLHttpRequest(); // there's engine.ajax
    xhr.open('post', '/api/stats/report', false);
    xhr.send(data);
  }
});
```

Here's the code I'd expect here. No timeouts, no intervals. No over-complication. [Source 1](https://w3c.github.io/beacon/), [Source 2](http://www.sitepoint.com/introduction-beacon-api/).

RobertoUa: don't I need it to do it like this ` var xhr = ('XMLHttpRequest' in root) ? new XMLHttpRequest() : new ActiveXObject('Microsoft.XMLHTTP');`

skavish: looks good except that we need to submit analytics by image get (as your comments reminded me)

PR #328, opened 2015-09-04
Configuration test page

shamansir:
An useful page to test different parameters for:

* Embedding with `IFRAME`
* Using Player from code
* Using Player as HTML element with `anm-player-target` attribute.

Will be improved, new options will be added, of course, but quite OK to use.


skavish:
ok, great! Iaroslav knows about it?

shamansir:
Yes, but I will also make a public announcement today.

PR #330, opened 2015-09-10
don't use setTimeout in analytics

RobertoUa:
@shamansir

### DIFF
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -4,7 +4,7 @@ var engine = require('engine'),
 var Analytics = function () {
     var self = this,
         supportSendBeacon = !!navigator.sendBeacon,
-        timeout = 1000,
+        interval = 1000,
         beacon = null,
         animatronUrl = utils.makeApiUrl('/analytics/player');
 
@@ -22,9 +22,6 @@ var Analytics = function () {
                 sendViaGif(data);
             }
         }
-        if (!supportSendBeacon) {
-            setTimeout(event, timeout);
-        }
     };
 
     var sendViaGif = function (data) {
@@ -35,7 +32,7 @@ var Analytics = function () {
     };
     if (self.enabled) {
         if (!supportSendBeacon) {
-            engine.onDocReady(event);
+            setInterval(event, interval);
         }
         window.addEventListener('unload', event, false);
     }
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -3,9 +3,6 @@ var engine = require('engine'),
 
 var Analytics = function () {
     var self = this,
-        supportSendBeacon = !!navigator.sendBeacon,
-        interval = 1000,
-        beacon = null,
         animatronUrl = utils.makeApiUrl('/analytics/player');
 
     self.enabled = animatronUrl != null && animatronUrl.indexOf('animatron-test') >= 0;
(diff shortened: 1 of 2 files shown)


skavish:
you forgot Dima

RobertoUa:
@skavish are you happy as well? :)

src/anm/analytics.js
```diff
@@ -16,27 +13,15 @@ var Analytics = function () {
...
             }
         }
-        if (!supportSendBeacon) {
-            setTimeout(event, timeout);
-        }
     };
 
-    var sendViaGif = function (data) {
-        if (!beacon) {
-            beacon = engine.createStatImg();
```
shamansir: Sorry for mentioning, but as I recall it was important to send it with image (:blush:) due to client-server issues (CORS?). But if you it's because of amount of data and/or you know what are you doing here, that's ok.

shamansir:
In general, it's great, as for me!

src/anm/analytics.js
RobertoUa: Well that was one of the reasons I went with the image, but you made me do this.
Anyways, if there's CORS issues then movie can't be loaded at all, since it is loaded via CORS. Am I right?

shamansir:
Thank you.

src/anm/analytics.js
shamansir: > You made me do this

No way :). An example is still just an example. 

> Anyways, if there's CORS issues then movie can't be loaded at all, since it is loaded via CORS. Am I right?

Dima @skavish should know.

skavish: the problem is not that the data won't be sent, but that there will be extra request: OPTIONS. no analytics software sends xhr requests because of that, everybody does it via image to minimize number of calls

RobertoUa: well now it will send data when page is closed.

shamansir: ...with XHR request. Please bring back `sendViaGif` and replace `engine.ajax` call with that. 

Is there any limit to amount of data that could be sent this way? May be we should optimize package size, since now we send it all at once.

Also, @RobertoUa, I think now [these options](https://github.com/Animatron/player/pull/330/files#diff-7f5481547366f17ae675b154345e696eR35) have no reason to be encrusted in every event record, they could be sent just once. To save data amount and time to send it.

RobertoUa: >sendViaGif and replace engine.ajax 

Then I will need to bring back `setInterval` since the image way won't work with `unload` event.

So all this arguing was useless?

skavish: i wonder how sendbeacon works on cross domain requests? if it also sends OPTIONS (i guess it should) then we cannon use it except on one case on window unload if we have anything to send

RobertoUa: As far as i know OPTIONS are not always sent. Usually when it has some extra header.

RobertoUa: Yes, I'm right. I just checked in browser - no OPTIONS. Also http://stackoverflow.com/a/13030629

RobertoUa: sendBeacon uses text/plain and POST. No custom headers.

skavish: ok, good then.

shamansir: Ok, good then. Let's test it in the action finally :).

> since the image way won't work with `unload` event.

I wasn't aware of that. Anyway, my point was is in unnecessary timeout calls, so no, not useless in the end ;).

PR #332, opened 2015-09-17
interactivity analytics

RobertoUa:
@shamansir like this

### DIFF
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -53,6 +53,20 @@ Analytics.prototype.trackPlayer = function trackPlayer(name) {
     }.bind(this);
 };
 
+Analytics.prototype.trackUI = function trackUI(player, path, type, time) {
+    var opts = {
+        viewId: player.viewId,
+        projectId: player.anim.meta._anm_id,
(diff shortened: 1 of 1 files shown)


src/anm/analytics.js
```diff
@@ -53,6 +53,20 @@ Analytics.prototype.trackPlayer = function trackPlayer(name) {
...
+    var opts = {
+        viewId: player.viewId,
+        projectId: player.anim.meta._anm_id,
+        time: player.state.time,
+        interactivity: {
+            path: path,
+            type: type
+        }
+    };
+    if (utils.is.num(time)) opts.interactivity.time = time;
```
shamansir: why not this: 

```javascript
var opts = {
    // ...
    time:  utils.is.num(time) ? time : player.state.time,
    // ...
}
```

shamansir: uh, it's different `time`, go it. why two of them, then? :)

RobertoUa: actually that's what I was going to do, but I forgot.

shamansir: ok, good!

PR #334, opened 2015-09-18
send user when sendBeacon is not supported

RobertoUa:
`withCredentials` wouldn't work.

Also it's probably not the best way to use regex for cookies :/

### DIFF
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -16,7 +16,8 @@ var Analytics = function () {
             if (navigator.sendBeacon) {
                 navigator.sendBeacon(animatronUrl, data);
             } else {
-                var auth = document.cookie.match(/_animatronauth=(\w+);/)[1];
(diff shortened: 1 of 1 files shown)


RobertoUa:
@shamansir

src/anm/analytics.js
```diff
@@ -16,7 +16,9 @@ var Analytics = function () {
             if (navigator.sendBeacon) {
                 navigator.sendBeacon(animatronUrl, data);
             } else {
-                engine.ajax(animatronUrl, null, null, 'POST', null, data, false);
+                var auth = document.cookie.match(/_animatronauth=(\w+);/)[1];
```
RobertoUa: maybe I should check for null and length > 0 before using `[1]`

shamansir: Yes, you should :). Also, final cookie may have no `;` sign.

RobertoUa:
updated

PR #336, opened 2015-10-06
Barna/analytics new format

RobertoUa:
added `trackPlayingComplete` on detach and also on page close/navigate.
new format, easier to aggregate on backend, and much faster
support http://analytics.animatron-test.com/

DO NOT MERGE until http://analytics.animatron-test.com/ is up and ready

### DIFF
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -55,6 +55,7 @@ Analytics.prototype.track = function track(name, player, action) {
                 screenWidth: screen.width,
                 windowHeight: window.innerHeight,
                 windowWidth: window.innerWidth,
+                timestamp: new Date().getTime(),
                 actions: []
             };
         action = action || {};
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -6,12 +6,23 @@ var Analytics = function () {
         animatronUrl = utils.makeApiUrl('/analytics/player');
 
     self.enabled = animatronUrl != null && animatronUrl.indexOf('animatron-test') >= 0;
-    self.queue = [];
+    self.queue = {};
 
-    var event = function () {
-        if (self.queue.length > 0) {
-            var data = JSON.stringify(self.queue);
-            self.queue = [];
+    var event = function (e) {
+        if (e && e.type === 'unload') {
+            anm.player_manager.instances.forEach(function (player) {
+                if (player && player.state && player.canvas) {
+                    self.trackPlayingComplete(player);
+                }
+            })
+        }
+        if (utils.is.not_empty(self.queue)) {
+            var array = [];
+            for (var key in self.queue) {
+                array.push(self.queue[key]);
+            }
+            self.queue = {};
+            var data = JSON.stringify(array);
 
             if (navigator.sendBeacon) {
                 navigator.sendBeacon(animatronUrl, data);
@@ -23,7 +34,6 @@ var Analytics = function () {
             }
         }
     };
-
     if (self.enabled) {
         window.addEventListener('unload', event, false);
     }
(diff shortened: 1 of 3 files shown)


RobertoUa:
@shamansir @skavish http://analytics.animatron-test.com/ is up and ready.(except for ssl support)
So it's ready to review and merge

src/anm/analytics.js
```diff
@@ -33,40 +43,41 @@ var Analytics = function () {
...
-    opts.url = location.href;
-    opts.screenHeight = screen.height;
-    opts.screenWidth = screen.width;
-    opts.windowHeight = window.innerHeight;
-    opts.windowWidth = window.innerWidth;
-    opts.timestamp = new Date().getTime();
+Analytics.prototype.track = function track(name, player, action) {
     if (this.enabled) {
-        this.queue.push(opts);
+        this.queue[player.viewId] = this.queue[player.viewId] || {
```
shamansir: It would be nicer to create some separate method for collecting this data (id's and screen metrics and everything else) first time for every player.

src/anm/analytics.js
```diff
@@ -33,40 +43,41 @@ var Analytics = function () {
...
+                viewId: player.viewId,
+                projectId: player.anim.meta._anm_id,
+                referer: document.referrer,
+                lang: navigator.language || navigator.userLanguage,
+                url: location.href,
+                screenHeight: screen.height,
+                screenWidth: screen.width,
+                windowHeight: window.innerHeight,
+                windowWidth: window.innerWidth,
```
shamansir: JFYI, this is a quite [slow operation](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) and causes layout reflow, it shouldn't occur during playing, but please keep this in mind.

src/anm/analytics.js
```diff
@@ -33,40 +43,41 @@ var Analytics = function () {
...
-    opts.screenWidth = screen.width;
-    opts.windowHeight = window.innerHeight;
-    opts.windowWidth = window.innerWidth;
-    opts.timestamp = new Date().getTime();
+Analytics.prototype.track = function track(name, player, action) {
     if (this.enabled) {
-        this.queue.push(opts);
+        this.queue[player.viewId] = this.queue[player.viewId] || {
+                viewId: player.viewId,
+                projectId: player.anim.meta._anm_id,
```
shamansir: Project ID could be changed for the same player. I mean a different project may be loaded into the same.

src/anm/analytics.js
```diff
@@ -3,15 +3,26 @@ var engine = require('engine'),
...
-            self.queue = [];
+    var event = function (e) {
+        if (e && e.type === 'unload') {
+            anm.player_manager.instances.forEach(function (player) {
+                if (player && player.state && player.canvas) {
+                    self.trackPlayingComplete(player);
+                }
+            })
+        }
+        if (utils.is.not_empty(self.queue)) {
```
shamansir: Why not sending an empty array if there's no data in `queue`?

src/anm/analytics.js
RobertoUa: yes I know. I causes relayout but it doesn't mean that it is slow(take a few ms, if not less). Because it happens only once, it's not in a loop or smth.

src/anm/analytics.js
RobertoUa: why would I? To avoid a control structure?

src/anm/analytics.js
RobertoUa: seriously? How come?

RobertoUa: then viewId should be changed too, no?

skavish: what is viewId?

RobertoUa: id of a view :)
I need some id so that i can group actions(play, pause, etc)

RobertoUa: generated on player instance creation

shamansir: > seriously? How come?

> generated on player instance creation

you answered this yourself. player may stop one project and and play another inside one instance.

shamansir: What I mean is `this.queue[player.viewId]` is filled with this object once for every player instance, so you set `projectId: player.anim.meta._anm_id` once for a player instance. But this very player could load another project and play it next to current one, and so on.

src/anm/analytics.js
shamansir: To make server know analytics were initialized, but no tracking events happened. Probably.

src/anm/analytics.js
shamansir: Also, regarding `viewId`, why you dislike `player.id` for that, it's [guaranteed to be unique](https://github.com/Animatron/player/blob/develop/src/anm/player.js#L696).

shamansir:
Ok, I merge it, just keep in mind the mentioned notes and make a new PR if you have new changes.

src/anm/analytics.js
shamansir: >  On a desktop browser, often the performance impact seems to be relatively unnoticeable. The user impact at a 10ms reflow time is rather negligible. However, herein seems to be much of the performance loss on mobile devices. The same pages that take only 10ms to reflow on a desktop can take over 1000ms on a mobile device. Every single change to the DOM further impacts that time (this is part of the reason why JavaScript-based animation looks so terrible on mobile devices).

I can't say player does no reflows at all, on the other hand, but if new ones can be avoided, why not do it one time for a page, not for a player?

PR #340, opened 2015-10-15
[PLAY-723] - base implementation

Ihromant:
[PLAY-723] - base implementation

### DIFF
--- a/src/anm/animation/element.js
+++ b/src/anm/animation/element.js
@@ -2901,7 +2901,7 @@ Element.getTranslate = function(elm) {
         return result;
     }
     return null;
-}
+};
 
 Element.getRotate = function(elm) {
     if (elm.parent.layer2Bone) {
@@ -2913,7 +2913,7 @@ Element.getRotate = function(elm) {
         return result;
     }
     return 0;
-}
+};
 
 Element.bonePath = function(elm) {
     var l2b = elm.parent.layer2Bone;
@@ -2923,7 +2923,7 @@ Element.bonePath = function(elm) {
        elm.$bonePath.add(bone);
        bone = l2b[bone.$from];
     }
-}
+};
 
 Element.getIMatrixOf = function(elm, m) {
     var t = Element.getMatrixOf(elm, m);
--- a/src/anm/animation/element.js
+++ b/src/anm/animation/element.js
@@ -2868,8 +2868,9 @@ Element.transferTime = function(src, trg) {
 Element.getMatrixOf = function(elm, m) {
     var t = (m ? (m.reset(), m)
                 : new Transform());
-    t.translate(elm.x, elm.y);
-    t.rotate(elm.angle);
+    var tr = Element.getTranslate(elm);
+    t.translate(tr ? tr.x : elm.x, tr ? tr.y : elm.y);
+    t.rotate(elm.angle + Element.getRotate(elm));
     t.shear(elm.hx, elm.hy);
     t.scale(elm.sx, elm.sy);
     t.translate(-elm.$reg[0], -elm.$reg[1]);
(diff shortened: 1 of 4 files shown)


src/anm/animation/element.js
```diff
@@ -2884,6 +2885,46 @@ Element.getMatrixOf = function(elm, m) {
...
+	elm.$bonePath = elm.$bonePath || Element.bonePath(elm);
+        var result = {x: 0; y: 0};
+        var bone = null;
+	for (var li = elm.$bonePath.length; li--;) {
+            bone = elm.$bonePath[li];
+	    result.x += bone.bonelength * Math.cos(bone.bonerotate);
+	    result.y += bone.bonelength * Math.sin(bone.bonerotate); 
+	}
+        elm = bone ? elm.parent.children[bone.$from] : elm;
+        result.x += elm.x;
```
shamansir: Actually these functions could be moved into bone tween(s) code, for example see [rotate-to-path](https://github.com/Animatron/player/blob/develop/src/anm/animation/tween.js#L263) implementation (and I'd personally prefer it this way) .

src/anm/animation/element.js
```diff
@@ -2884,6 +2885,46 @@ Element.getMatrixOf = function(elm, m) {
...
+}
+
+Element.getRotate = function(elm) {
+    if (elm.parent.layer2Bone) {
+        elm.$bonePath = elm.$bonePath || Element.bonePath(elm);
+	var result = 0;
+	for (var li = elm.$bonePath.length; li--;) {
+	    result += elm.$bonePath[li].bonerotate;
+	}
+        return result;
```
shamansir: Same here.

src/anm/animation/tween.js
```diff
@@ -254,6 +254,22 @@ Tween.register(C.T_ROTATE, function(values) {
...
+    return function(t) {
+        this.angle = _from * (1.0 - t) + to * t;
+    }
+});
+
+Tween.register(C.T_BONE_LENGTH, function(values) {
+    var _from = values[0],
+        to = values[1];
+    return function(t) {
+        this.angle = _from * (1.0 - t) + to * t;
```
shamansir: You may safely do `+=` in these tweens. See [rotate-to-path tween](https://github.com/Animatron/player/blob/develop/src/anm/animation/tween.js#L263) implementation. But also please notice that tweens like this one should exist during the whole lifetime of an element: that's why [here](https://github.com/Animatron/player/blob/develop/src/import/animatron-importer.js#L302) we add a rotate tween with first value of an angle at start and a tween with final value of an angle in the end.

shamansir:
Please notice tests do fail (while there's not a lot of tests and they are unrelated), you may run them locally with `jake test`.

src/anm/animation/element.js
Ihromant: It's harder to do this because boneLength and boneRotate have influence to alot of parts of skeleton, not the one layer

shamansir: I think if this tween will be assigned to the every child and this child will check its parent, this could work. But `elm.x` will be taken from a previous frame, is that ok?

src/anm/animation/element.js
```diff
@@ -2884,6 +2885,46 @@ Element.getMatrixOf = function(elm, m) {
     return t;
 };
 
+Element.getTranslate = function(elm) {
```
shamansir: By the way, you may do it `Element.prototype.getTranslate`, then it will be instance method, not static method, and element will be accessible as `this`.

shamansir:
Just for everyone's information: 

This PR is OK and works. I proposed to postpone it a bit, not to mix with new timing changes.
Secondarily, it needs a fix in rendering Preview and also I asked Ivan to try to find a way to isolate bone logic inside tweens, instead of using `Element.getTranslate`/`Element.getRotate`. If that's possible.

Ihromant:
Actually finished this. There is an issue with order of layers in skeleton, I created task for it:
http://youtrack.animatron.com/youtrack/issue/PLAY-760

shamansir:
I may merge it, but I created a task for myself regarding encapsulating bone logic inside bone tweens: http://youtrack.animatron.com/youtrack/issue/PLAY-761, just to have it in mind. High chances it will be forgotten, though...

shamansir:
Thank you!

PR #341, opened 2015-10-15
handle video size in exported clip

skavish:


skavish:
I realized we need to export size of the video rectangle as well, we cannot rely on video size because it could be different from the original size or even from one format to another due to downsampling/size reducing/etc processing we do.

I updated compact format as well, consult it if you need.

shamansir:
Looks good, thanks!

PR #342, opened 2015-10-19
Send analytic data on player detach

RobertoUa:
@shamansir @skavish pls review

### DIFF
--- a/src/anm/analytics.js
+++ b/src/anm/analytics.js
@@ -1,14 +1,15 @@
-var engine = require('engine'),
-    utils = require('./utils.js');
+var engine = require('engine');
+var utils = require('./utils.js');
+var C = require('./constants.js');
 
 var Analytics = function () {
-    var self = this,
-        animatronUrl = utils.makeApiUrl('analytics', '/analytics/player');
+    var self = this;
+    self.animatronUrl = utils.makeApiUrl('analytics', '/analytics/player');
 
     self.enabled = animatronUrl != null && animatronUrl.indexOf('animatron-test') >= 0;
     self.queue = {};
 
-    var event = function (e) {
+    var sendAllToServer = function (e) {
         if (e && e.type === 'unload') {
             anm.player_manager.instances.forEach(function (player) {
                 if (player && player.state && player.canvas) {
@@ -19,23 +20,17 @@ var Analytics = function () {
         if (utils.is.not_empty(self.queue)) {
             var array = [];
             for (var key in self.queue) {
-                array.push(self.queue[key]);
+                if (self.queue.hasOwnProperty(key)) {
+                    array.push(self.queue[key]);
+                }
             }
             self.queue = {};
-            var data = JSON.stringify(array);
-
-            if (navigator.sendBeacon) {
-                navigator.sendBeacon(animatronUrl, data);
-            } else {
-                var match = document.cookie.match(/_animatronauth=([^;]+)/);
-                var auth = match && match.length > 0 ? match[1] : '';
-                var params = auth ? '?user=' + auth : '';
-                engine.ajax(animatronUrl + params, null, null, 'POST', null, data, false);
-            }
+            self.sendData(array)
         }
     };
+
     if (self.enabled) {
-        window.addEventListener('unload', event, false);
+        window.addEventListener('unload', sendAllToServer, false);
     }
 
     this.trackPlayingStart = this.trackPlayer('playing_start');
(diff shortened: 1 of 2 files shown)


src/anm/analytics.js
```diff
@@ -70,6 +84,13 @@ Analytics.prototype.trackPlayer = function trackPlayer(name) {
     return function (player) { this.track(name, player); }.bind(this);
 };
 
+Analytics.prototype.sendPlayerData = function sendForPlayer(player) {
```
shamansir: what is `sendForPlayer ` here?

src/anm/analytics.js
```diff
@@ -43,6 +37,26 @@ var Analytics = function () {
     this.trackPlayingComplete = this.trackPlayer('playing_complete');
 };
 
+/**
+ * @param {Array} views - an array of views
+ */
+Analytics.prototype.sendData = function sendData(views) {
```
shamansir: I think having two methods with similar names `sendData`/`sendPlayerData` is not very good, either one of them could be private (marked with `@private`) or they should have different names.

shamansir: Also, `sendData` is written twice here.

src/anm/analytics.js
```diff
@@ -43,6 +37,26 @@ var Analytics = function () {
...
+        engine.ajax(this.animatronUrl + params, null, null, 'POST', null, data, false);
+    }
+};
+
+/**
+ * @param {String} name
+ * @param {Player} player
+ * @param {Object} [action]
+ */
 Analytics.prototype.track = function track(name, player, action) {
```
shamansir: And `track` here, btw.

src/anm/analytics.js
```diff
@@ -19,23 +19,17 @@ var Analytics = function () {
         if (utils.is.not_empty(self.queue)) {
             var array = [];
             for (var key in self.queue) {
-                array.push(self.queue[key]);
+                if (self.queue.hasOwnProperty(key)) {
```
shamansir: Since you use `forEach` (ES5+) above anyway, I think you also may use smth like 

```javascript
var array = Object.keys(self.queue).map(function (key) { return self.queue[key]; });
```

src/anm/analytics.js
```diff
@@ -70,6 +84,13 @@ Analytics.prototype.trackPlayer = function trackPlayer(name) {
     return function (player) { this.track(name, player); }.bind(this);
 };
 
+Analytics.prototype.sendPlayerData = function sendForPlayer(player) {
+    this.trackPlayingComplete(player);
+    var action = [this.queue[player.viewId]];
+    this.queue[player.viewId] = undefined;
+    this.sendData(action)
```
shamansir: Moving data from objects to arrays and back looks weird, actually, could this be changed to managing one data type? If you always send updates one by one, why do you need arrays? Here, anyway, you clear the hash cell for this `viewId`.

shamansir:
Please fix function declarations at least.

src/anm/analytics.js
RobertoUa: oh, I overlooked this, thank you!

src/anm/analytics.js
RobertoUa: I did this because I didn't want to use anonymous function.

shamansir:
That's not right:

```javascript
Analytics.prototype.sendPlayerData = function sendPlayerData(player) {
```

should be:

```javascript
Analytics.prototype.sendPlayerData = function(player) {
```

Same [here](https://github.com/Animatron/player/pull/342/files#diff-7f5481547366f17ae675b154345e696eR43) and [here](https://github.com/Animatron/player/pull/342/files#diff-7f5481547366f17ae675b154345e696eR60)

src/anm/analytics.js
RobertoUa: > always send updates one by one
not always.
>Moving data from objects to arrays and back looks weird
Server requires an array(map doesn't fit here).

src/anm/analytics.js
RobertoUa: Right, I shouldn't have used `forEach`. But here you say that it's ok. Are you sure that we don't need to support old browsers here?

RobertoUa:
Since when it's not "right". It's just a matter of style. If you use anonymous functions (as you suggest) you will notice `(anonymous function)` in stacktraces. If you want me to follow your code style, it's ok, I don't really care which one to use.

shamansir:
I think that's a common style not to type a function name twice. Especially when it's only has a reason in making debugging easier.

shamansir:
Thanks!

PR #344, opened 2015-10-19
[PLAY-753] fixed. Video does not play

RobertoUa:
@shamansir @skavish

### DIFF
--- a/src/anm/media/video.js
+++ b/src/anm/media/video.js
@@ -137,7 +137,7 @@ Video.prototype.load = function(elm, player) {
 };
 /** @private @method apply */
 Video.prototype.apply = function(ctx) {
-    if (this.video) ctx.drawImage(this.video, 0, 0, this.video.width, this.video.height, 0, 0, this.size[0], this.size[1]);
+    if (this.video) ctx.drawImage(this.video, 0, 0, this.video.videoWidth, this.video.videoHeight, 0, 0, this.size[0], this.size[1]);
 };
 Video.prototype.bounds = function() {
     if (this.$bounds) return this.$bounds;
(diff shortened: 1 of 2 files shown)


src/engine/dom-engine.js
```diff
@@ -924,7 +924,9 @@ $DE.createAudio = function() {
 };
 
 $DE.createVideo = function() {
-    return $doc.createElement('video');
+    var el = $doc.createElement('video');
+    el.width = el.height = "1";
```
shamansir: shouldn't it be `1px` at least?

RobertoUa:
updated

src/engine/dom-engine.js
RobertoUa: no

shamansir:
ok

RobertoUa:
updated for videos larger than 720

shamansir:
@skavish do you approve?

src/anm/media/video.js
```diff
@@ -137,7 +137,7 @@ Video.prototype.load = function(elm, player) {
 };
 /** @private @method apply */
 Video.prototype.apply = function(ctx) {
-    if (this.video) ctx.drawImage(this.video, 0, 0, this.video.width, this.video.height, 0, 0, this.size[0], this.size[1]);
+    if (this.video) ctx.drawImage(this.video, 0, 0, this.video.videoWidth, this.video.videoHeight, 0, 0, this.size[0], this.size[1]);
```
skavish: yep, that's correct fix, at least those are always be set if the video is loaded.

I was contemplating what to use in the editor: width or videoWidth and for whatever reason chose width, although videoWidth is better

PR #345, opened 2015-10-20
fix exception for video when playing locally

RobertoUa:
@shamansir for video url is array.

### DIFF
--- a/src/anm/media/video.js
+++ b/src/anm/media/video.js
@@ -47,7 +47,7 @@ Video.prototype.load = function(elm, player) {
     var me = this;
     ResMan.loadOrGet(player.id, me.url,
         function(notify_success, notify_error, notify_progress) { // loader
-            var url = me.url;
+            var url = Array.isArray(me.url) && me.url.length > 0 ? me.url[0] : me.url;
             var formats = me.formats;
             if (engine.isHttps) { url = url.replace('http:', 'https:'); }
             url = engine.fixLocalUrl(url);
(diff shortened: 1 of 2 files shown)


RobertoUa:
I don't know whether I do it as it's suppose to be done but at least now it works.
@skavish

src/anm/media/video.js
```diff
@@ -47,7 +47,7 @@ Video.prototype.load = function(elm, player) {
     var me = this;
     ResMan.loadOrGet(player.id, me.url,
         function(notify_success, notify_error, notify_progress) { // loader
-            var url = me.url;
+            var url = Array.isArray(me.url) && me.url.length > 0 ? me.url[0] : me.url;
```
shamansir: Could you please use a bit more parentheses to group expressions in code, they really help visually. Yes, even when operator precedence is obvious.

shamansir:
Actually it shouldn't be an array, the format URL [used as sources](https://github.com/RobertoUa/player/blob/barna/fix_local_url_for_video_exception/src/anm/media/video.js#L118) should be an array. So I'd prefer to find a reason why it's an array.

src/anm/media/video.js
RobertoUa: Frankly, for me extra symbols make it harder to read code in such simple cases. But since it's your codebase, I will follow your code style.

RobertoUa:
Maybe, let's wait @dima, because it happened after his changes to editor

src/anm/media/video.js
shamansir: Thanks! `Array.isArray(me.url) && (me.url.length > 0) ? me.url[0] : me.url` would be fine in this case, no need in extremes, of course.

shamansir:
ok.

src/anm/media/video.js
skavish: sorry, it was my fault. somehow this array crept in, funny thing I remember that a test failed, but I just went and "fixed" it by supplying wrong data thinking it's correct.

I will fix it on editor side, so no need for this PR

PR #348, opened 2015-10-22
Add animatron.local.js bundle without analytics

RobertoUa:
@shamansir pls review

### DIFF
--- a/src/anm/utils.js
+++ b/src/anm/utils.js
@@ -277,6 +277,13 @@ function makeApiUrl(prefix, path, loadSrc) {
     }
 }
 
+function getObjectId () {
+    var timestamp = (new Date().getTime() / 1000 | 0).toString(16);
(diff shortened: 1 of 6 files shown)


src/anm/utils.js
```diff
@@ -277,6 +277,13 @@ function makeApiUrl(prefix, path, loadSrc) {
     }
 }
 
+function getObjectId () {
```
shamansir: what's the major difference with `utils.guid`, more reliable random? :)

shamansir:
Looks great in general, nice! Thank you!

src/anm/utils.js
RobertoUa: It's not about random, it's about ObjectId.

shamansir: You mean it's compliant to a spec? Ok, got it.

PR #352, opened 2015-11-02
Refactor timing

shamansir:
A lot of things changed with timing, there's a separate `Timeline` class now and it works completely different to what was before (in the core it uses time increment instead of bands, so it could call time-based handlers and decide if Element continues living from the inside).

Also, it has `Scene` class now, scenes are behaving a bit differently to elements — they're sequenced in one line of the time, so when first scene reached the end, the next one comes.


spec/search.spec.js
```diff
@@ -105,8 +114,8 @@ describe('search', function() {
             animation.add(element('stub'));
             animation.add(searchFor);
 
-            expect(animation.find('/foobar')).toBe(searchFor);
-            expect(animation.findAll('/foobar')[0]).toBe(searchFor);
+            expect(animation.find('//foobar')).toBe(searchFor);
```
skavish: is it going to break existing projects?

shamansir: No, there's one default Scene (named `""`) in any Animation now (it guarantees there's always one scene), but on import from Animatron this scene is replaced with a first scene of a project, so `/Scene 1/Layer 1` should work still.

shamansir: ( [The line](https://github.com/Animatron/player/pull/352/files#diff-450f96d427a7e52e0cc876bee087c640R95) where it happens on import. )

PR #359, opened 2015-11-13
Ihetman/play 723impr

Ihromant:
Fixed some bugs with misplacement and code refactoring

### DIFF
--- a/src/anm/animation/element.js
+++ b/src/anm/animation/element.js
@@ -2553,9 +2553,9 @@ Element.transferTime = function(src, trg) {
 Element.getMatrixOf = function(elm, m) {
     var t = (m ? (m.reset(), m)
                 : new Transform());
-    var tr = Element.getTranslate(elm);
-    t.translate(tr ? tr.x : elm.x, tr ? tr.y : elm.y);
-    t.rotate(elm.angle + Element.getRotate(elm));
+    var tr = elm.getTranslate();
+    t.translate(tr.x, tr.y);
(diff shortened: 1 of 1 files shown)


src/anm/animation/element.js
```diff
@@ -2570,45 +2570,50 @@ Element.getMatrixOf = function(elm, m) {
     return t;
 };
 
-Element.getTranslate = function(elm) {
-    if (elm.parent && elm.parent.layer2Bone) {
-	elm.$bonePath = elm.$bonePath || Element.bonePath(elm);
+Element.prototype.getTranslate = function() {
+    if (this.parent && this.parent.layer2Bone) {
```
shamansir: I still hope it will be a tween at some day :)

src/anm/animation/element.js
```diff
@@ -2570,45 +2570,50 @@ Element.getMatrixOf = function(elm, m) {
...
-    var result = [];
-    var bone = l2b[elm.parent.children.indexOf(elm)];
-    while (bone) {
-       result.push(bone);
-       bone = l2b[bone.$from];
+Element.prototype.bonePath = function() {
+    if (!this.$bonePath) {
+        var l2b = this.parent.layer2Bone;
+        var result = [];
+        var bone = l2b[this.parent.children.indexOf(this)];
```
shamansir: Each element has an unique ID in `elm.id`, so you may use it to store in `layer2Bone`, instead of index, which actually could change in runtime in some rare cases.

PR #360, opened 2015-11-13
Animation and Scene now also contain Timeline and Player interacts with it

shamansir:
+ All of the rendering is now based on `dt` and timeline positions instead of global time;
+ Improved controls;


shamansir:
Yet needs some tests, please do not merge.

PR #364, opened 2015-12-08
[PLAY-770] fixed. Google fonts should be supported by the player

vkoresh:
@shamansir please review.

### DIFF
--- a/src/anm/animation/animation.js
+++ b/src/anm/animation/animation.js
@@ -587,7 +587,13 @@ Animation.prototype._collectRemoteResources = function(player) {
         }
     });
     if (this.fonts && this.fonts.length) {
-        remotes = remotes.concat(this.fonts.map(function(f){return f.url;}));
+        remotes = remotes.concat(this.fonts.map(function(f){
+            if (!f.gf_name) {
+                return f.url;
+            } else {
+                return null;
+            }
+        }));
     }
     return remotes;
 };
@@ -721,6 +727,7 @@ var FONT_LOAD_TIMEOUT = 10000; //in ms
  * @method loadFonts
  * @private
  */
+var URL2FG = 'http://fonts.googleapis.com/css?family=';
 Animation.prototype.loadFonts = function(player) {
     if (!this.fonts || !this.fonts.length) {
         return;
@@ -731,7 +738,7 @@ Animation.prototype.loadFonts = function(player) {
         css = '',
         fontsToLoad = [],
         detector = new FontDetector();
-
+    var url2gf = '';
     for (var i = 0; i < fonts.length; i++) {
         var font = fonts[i];
         if (!font.url || !font.face) {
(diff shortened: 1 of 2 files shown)


src/anm/animation/animation.js
```diff
@@ -721,6 +727,7 @@ var FONT_LOAD_TIMEOUT = 10000; //in ms
  * @method loadFonts
  * @private
  */
+var URL2FG = 'http://fonts.googleapis.com/css?family=';
```
shamansir: Probably that one should be `URL2GF`? But I'd prefer some less abbreviated name anyway :)

src/engine/dom-engine.js
```diff
@@ -906,6 +906,20 @@ $DE.createStatImg = function() {
...
 
+$DE.addFontLinkObject = function(url) {
+    var link = document.getElementById('anm-link');
+    if (link) {
+        return link;
+    }
+    link = $doc.createElement('link');
+    link.type = 'text/css';
+    link.id = 'anm-link';
+    link.rel = "stylesheet";
```
shamansir: Just a code style issue, single-quote strings are preferred in the Player code.

src/anm/animation/animation.js
```diff
@@ -739,21 +746,30 @@ Animation.prototype.loadFonts = function(player) {
...
+            (font.style ? 'font-style: ' + font.style + ';\n' : '') +
             (font.weight ? 'font-weight: ' + font.weight + ';\n' : '') +
             '}\n';
+        } else {
+            url2gf += font.gf_name + "|";
+        }
+    }
+    if (url2gf != '') {
+        var link = URL2FG + url2gf.substring(0, url2gf.lastIndexOf('|'));
+        console.log("inserting link for google fonts: " + link);
```
shamansir: Please use `var log = require('../log.js')` for logging, if it's required.

src/anm/animation/animation.js
vkoresh: yes it is typo, thank you

src/anm/animation/animation.js
```diff
@@ -739,21 +746,30 @@ Animation.prototype.loadFonts = function(player) {
...
+            'src:' + (woff ? ' url("' + woff + '") format("woff"),\n' : '') +
+            ' url("' + url + '") format("truetype");\n' +
+            (font.style ? 'font-style: ' + font.style + ';\n' : '') +
             (font.weight ? 'font-weight: ' + font.weight + ';\n' : '') +
             '}\n';
+        } else {
+            url2gf += font.gf_name + "|";
+        }
+    }
+    if (url2gf != '') {
```
shamansir: it's better to use `if (url2gf.length) { ... }` to check if string is empty, or else you compare with instance of an empty string.

src/engine/dom-engine.js
```diff
@@ -906,6 +906,20 @@ $DE.createStatImg = function() {
     return img;
 };
 
+$DE.addFontLinkObject = function(url) {
+    var link = document.getElementById('anm-link');
```
shamansir: Please also use `$doc` here and below instead of `document`. That's not very relevant for a truly long time, but it's the established code of conduct :).

src/anm/animation/animation.js
```diff
@@ -721,6 +727,7 @@ var FONT_LOAD_TIMEOUT = 10000; //in ms
  * @method loadFonts
  * @private
  */
+var URL2GOOGLE_FONTS = 'http://fonts.googleapis.com/css?family=';
```
shamansir: nice!

vkoresh:
@shamansir updated

shamansir:
great, thank you a lot!

vkoresh:
np :)

skavish:
I thought you were going to use google font loader. if you don't use it here, then why do we need to use it in the editor?

shamansir:
Hmm, though I haven't discussed that with @vkoresh particularly, but I'd prefer *not* to use GFL and stay as vanilla as possible, for the sake of the Player "package" size. Also, adding one more remote alien resource loader would require more work to integrate it with Player's own Resource Manager (even while the latter is actually cache/subscribe manager). 

Yet if GFL provides _truly_ a lot of speed over this solution, I'd load/add GPL only for projects with Google Fonts, of course.

PR #366, opened 2015-12-17
https url to google fonts

vkoresh:
@shamansir pls review and merge

### DIFF
--- a/src/anm/animation/animation.js
+++ b/src/anm/animation/animation.js
@@ -727,7 +727,7 @@ var FONT_LOAD_TIMEOUT = 10000; //in ms
  * @method loadFonts
  * @private
  */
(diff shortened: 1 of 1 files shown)


shamansir:
I suppose, same way as for other media, it should use `http` when player was opened using `http`, and `https` if not. You may use [`engine.checkMediaUrl`](https://github.com/Animatron/player/blob/134bdcb986a0bbeabe9ced5abfc1a6eb8df313f1/src/engine/dom-engine.js#L965) for that: [examples](https://github.com/Animatron/player/search?utf8=%E2%9C%93&q=checkmediaurl).

vkoresh:
thanks! updated

shamansir:
Thank you!

PR #367, opened 2015-12-17
Refactor timing: Code review issues, masks refactoring

shamansir:


shamansir:
@skavish please take a look

src/anm/animation/element.js
```diff
@@ -805,14 +813,13 @@ Element.prototype.render = function(ctx) {
             this.transform(bctx);
             this.painters(bctx);
             this.each(function(child) {
-                child.tick(dt);
                 child.render(bctx, dt);
             });
 
+            // FIXME: this should be performed one time for all masked elements!
             mask.transform(mctx);
```
shamansir: @skavish I am uncertain in this part. The matrix is calculated once, but current bounds affect the mask canvas size. Is it safe to move it to mask render cycle?

src/anm/animation/element.js
```diff
@@ -2424,7 +2430,8 @@ Element.prototype.__adaptModTime = function(modifier, ltime) {
 Element.prototype.__pbefore = function(ctx, type) { };
 Element.prototype.__pafter = function(ctx, type) { };
 Element.prototype.filterEvent = function(type, evt) {
-    if (this.shown) this.__saveEvt(type, evt);
+    /*if (this.visible && this.active)*/
+    this.__saveEvt(type, evt);
```
shamansir: For the moment, cached events (like mouse events) are handled strictly before the tweens on the next frame and then cache is cleared. This technique probably outdated and may be removed.

src/anm/animation/animation.js
```diff
@@ -225,9 +225,9 @@ Animation.prototype.traverseVisible = function(visitor, data) {
  * @param {Object} [data]
  */
 Animation.prototype.reverseTraverseVisible = function(visitor, data) {
-    if (this.currentScene && this.currentScene.time.fits()) {
+    if (this.currentScene && this.currentScene.timeline.isActive()) {
```
skavish: above you just used currentScene.isActive

src/anm/animation/animation.js
```diff
@@ -225,9 +225,9 @@ Animation.prototype.traverseVisible = function(visitor, data) {
...
         this.currentScene.reverseTraverse(function(child) {
-            return (child.time.fits() && (visitor(child, data) === false)) ? false : true;
+            return (child.timeline.isActive() && (visitor(child, data) === false)) ? false : true;
```
skavish: same here

shamansir: missed that, thanks!

src/anm/animation/element.js
skavish: mask canvas size does not change during one rendering cycle. the masked elements and the mask itself are rendered into canvases the same size.

you need to draw all masked elements into a canvas first, then (after all masked elements are rendered) you need to render mask itself

shamansir: I've made a draft for this, not yet tested: https://github.com/Animatron/player/compare/refactor-timing...refactor-masks. I will ensure if order of rendering (masked elements before mask itself) is right later, but may be you'll see some pitfalls obvious for you.

skavish: I want to comment there, but I cannot comment on a diff, only on pullrequest

shamansir: You are welcome: https://github.com/Animatron/player/pull/368 :)

PR #368, opened 2015-12-17
Refactor masks, in progress

shamansir:


src/anm/animation/element.js
```diff
@@ -744,101 +744,141 @@ Element.prototype._checkSwitcher = function(next) {
...
+    if (this.disabled || !this.active) return;
 
     this.rendering = true;
 
     if (this.visible) {
-        ctx.save();
 
-        if (!this.$mask) {
-            // draw directly to context, if has no mask
+        if (!this.isMask && !this.$mask) {
```
skavish: it would be helpful to rename those variable like: isMask and isMasked

src/anm/animation/element.js
```diff
@@ -744,101 +744,141 @@ Element.prototype._checkSwitcher = function(next) {
...
 
-            mctx.restore();
-            bctx.restore();
-        }
-        ctx.restore();
+    var minX = Number.MAX_VALUE, minY = Number.MAX_VALUE,
+        maxX = Number.MIN_VALUE, maxY = Number.MIN_VALUE;
+
+    var pt;
+    for (var i = 0, il = bounds_pts.length; i < il; i++) {
```
skavish: what does it do exactly?

shamansir: if masking element is rotated, mask bounds should fit it completely (like "bounding rectangle"). It could be an artefact now, but I am not sure yet. See [`.bounds`](https://github.com/Animatron/player/blob/refactor-masks/src/anm/animation/element.js#L2002) and [`.adaptBounds`](https://github.com/Animatron/player/blob/refactor-masks/src/anm/animation/element.js#L2106).

PR #371, opened 2015-12-25
Switch tween fixes

shamansir:


src/anm/animation/element.js
```diff
@@ -709,24 +709,26 @@ Element.prototype.tick = function(dt) {
     }
     var isActive = this.timeline.isActive() && this.modifiers(resultTime, dt);
     if (isActive) {
-        this.each(function(child) {
-            child.tick(dt);
-        });
+        if (this.hasSwitch) {
+            this.each(function(child) {
+                if (child.name == this.switch) {
+                    if (this.justSwitched/* || ((this.getTime() - this.switchBand[0]) > child.timeline.getDuration())*/) child.timeline.reset();
```
shamansir: @skavish It is required to reset child time _once_ every time we overflow it's own lifetime duration. I haven't found nice way to do it for the moment :(

shamansir:
@skavish This one is OK to merge, but it still has some issues when switch band is larger than element band (see [comment](https://github.com/Animatron/player/pull/371#discussion_r48480823))

PR #375, opened 2016-01-14
Bubbling up all events except mouseenter/mouseexit

shamansir:


src/anm/events.js
```diff
@@ -152,26 +152,28 @@ MouseEventsSupport.prototype.adaptEvent = function(event) {
 MouseEventsSupport.prototype.dispatch = function(event) {
     var owner = this.owner;
```
shamansir: @kirillk @skavish please review changes below.

`dispatchedChild` is now the child instance, instead of boolean `dispatchedByChild`.

spec/mouse.spec.js
```diff
@@ -184,83 +185,110 @@ describe('handling mouse in static objects', function() {
...
+        var log = new EventLog(targets, events);
+
+        beforeEach(function() {
+            log.subscribe();
+        });
+
+        afterEach(function() {
+            log.unsubscribe();
+            log.clear();
+        });
```
shamansir: @kirillk @skavish Also please review tests below. Should event target be equal to the element received the event or the target (and event instance?) should be the same, like always `e11` here?

src/anm/events.js
```diff
@@ -152,26 +152,28 @@ MouseEventsSupport.prototype.adaptEvent = function(event) {
...
-        if (dispatchedByChild) return true;
-
-        if (event.type === 'mouseclick') {
-            this.owner.fire('mouseclick', localEvent);
-            return true;
-        } else if (event.type === 'mousemove') {
-            this.processMove(localEvent); // fire mouseenter/mousexit if required
-            return true;
+        if (!dispatchedChild && (event.type === 'mousemove')) {
+            this.processMove(localEvent); // fire mouseenter/mouseexit if required
```
skavish: should we still fire mousemove if mouseenter/exit were fired and handled? I am not sure, I'd rather not fire it in this case

shamansir: I think we should fire both at least for the `mouseenter` case—if you listen only for `mousemove`, then you need this first event too. But for `mouseexit` it's controversial, since it will have a `mouseenter` pair in some cases, in some cases it won't.

spec/mouse.spec.js
kirillk: I believe we should drop this "target", there's no need for it now, we're not pumping events through like, event bus or somethings

Let's make it barely sufficient for now

spec/mouse.spec.js
```diff
@@ -184,83 +185,110 @@ describe('handling mouse in static objects', function() {
...
 
-            log = [];
+            log.clear();
 
             fireCanvasEvent('mousemove', 25, 75);
-            expect(log.join(MARKER)).toEqual('e12: mouseexit@null;null -> e12' + MARKER +
-                                             'e1: mouseexit@null;null -> e12' + MARKER +
-                                             'e2: mouseenter@25;30 -> e2');
+            expect(log.stringify(MARKER)).toEqual([ 'e12: mouseexit@null;null -> e12',
+                                                    'e1: mouseexit@null;null -> e12',
```
kirillk: why do you have point on enter and don't have point on exit? I believe there shouldn't be any point on both enter/exit

again, we can compute it, but let;s implement everything gradually keeping the api simple

kirillk:
@shamansir I just realized that you made the tests different from the spec. Can't you see the difference?

``` public void testPress() {
        assertDispatchPress(
                25, 25,
                "e11: press@25,25"
        );

        assertDispatchPress(
                75, 25,
                "e1: press@75,25"
        );

        assertDispatchPress(
                25, 47,
                "e2: press@25,2"
        );

        assertDispatchPress(
                75, 47,
                "e2: press@75,2"
        );
    }```

**Comment by kirillk** [comment]
```  it('properly handles mousedown event', function() {
             fireCanvasEvent('mousedown', 10, 10);
             expect(log.stringify(MARKER)).toEqual([ 'e11: mousedown@10;10 -> e11',
                                                     'e1: mousedown@10;10 -> e1',
                                                     'root: mousedown@10;10 -> root' ].join(MARKER));
         });
```

kirillk:
@shamansir I'm adding listeners and checking events don't get bubbled. You're testing some internal processes not the use API

can you get the difference?

spec/mouse.spec.js
shamansir: Because user may be interested where is the mouse position of first enter, or else she should extract it from mousemove and determine which mousemove was first. On exit, position is not important and it's hard to determine which element it should be relative to, though we could have both for consistency.

kirillk: No such consistency needed. Enter might be not mean that the mouse is
actually over the element, it might be on a child. Please remove it.

shamansir:
@kirillk API is tested when I [subscribe for events here](https://github.com/Animatron/player/blob/develop/spec/mouse.spec.js#L315). This `.on` method is the same you call from scripts. I also emulate real canvas events or else I can't be sure they reach the API.

kirillk:
so this test case doesn't show us that events don't get bubbled, isn't that obvious?

isn't that clear that the test case is DIFFERENT from what's on the spec?

shamansir:
Uh, you are about clicks tests. Not that obvious, but I got it now, will fix it.

shamansir:
@kirillk Also we decided click should bubble, aren't we?

shamansir:
https://github.com/Animatron/player/commit/460812f36d4c6580caab5d41874ba56edb4f866d

@kirillk Please see this commit, there are updated tests, but they fail now, since clicks *do* bubble, as we decided ("All mouse events except mouseexit/mouseenter are bubbling").

shamansir:
@kirillk Just in case, it subscribes to all of the elements (`root`, `e1`, `e11`, `e12`, `e2`), for the given events (`mouseclick` in this case), same way your code does, so it actually checks bubbling now.

kirillk:
@shamansir what we decided on is in the spec. I have the tests and the ref impl for you. I don't know how it should be even more clear than this. Should I implement that myself?

spec/mouse.spec.js
```diff
@@ -145,38 +142,52 @@ describe('handling mouse in static objects', function() {
...
 
-            expect({ type: 'click', x: 25, y: 47 })
-               .toBeHandledAs({ type: 'mouseclick', target: e2, x: 25, y: 2 });
+            fireCanvasEvent('click', 25, 47);
+            expect(log.stringify(MARKER)).toEqual([ 'e2: mouseclick@25;2 -> e2' ].join(MARKER));
 
-            expect({ type: 'click', x: 75, y: 47 })
-               .toBeHandledAs({ type: 'mouseclick', target: e2, x: 75, y: 2 });
+            fireCanvasEvent('click', 75, 47);
+            expect(log.stringify(MARKER)).toEqual([ 'e2: mouseclick@25;2 -> e2' ].join(MARKER));
```
skavish: Anton, in Kirill's test it should be 75,2, why did you change it? it was right

skavish:
just to make it clear once and for all:

when we say events do bubble we mean that they bubble until we find a handler which handles it. as soon as the handler is found the bubbling should be stopped.

src/anm/events.js
```diff
@@ -152,26 +152,28 @@ MouseEventsSupport.prototype.adaptEvent = function(event) {
...
         }
+
+        //var adaptedEvent = localEvent.clone();
+        //adaptedEvent.target = dispatchedChild || this.owner;
+        //this.owner.fire(event.type, adaptedEvent);
+
+        this.owner.fire(event.type, localEvent);
```
skavish: it should be fired ONLY if none of the children handled it, i.e. 
```
   if (!dispatchedChild) {
       this.owner.fire(event.type, localEvent);
   }
```

also Anton, please have the tests to be exactly the same as in Kirill's tests. if they fail it means your implementation is incorrect and you need to fix the implementation, not change the tests!

shamansir:
@skavish Then it's not bubbling, please find the appropriate term. Same way, you've probably meant "dispatched" and not "handled", since "handled" means handler (listener) for an event was called.

shamansir:
@kirillk if you want something to be clear, please add detailed comments in your code. And yes, please, if you want, you are always welcome to participate. Also you may wait a bit till AI will be able to write code, machines are perfect in copying things.

skavish:
@shamansir we were discussing it for many hours and always talked about it as "bubbling with automatic stop propagation on handle", because that's what it is. this fact that the bubbling should be stopped on handle was stressed many times and by you as well.

I have no idea why you all of sudden decided that it should be different now and even changed the tests to accomodate it.

shamansir:
@skavish http://javascript.info/tutorial/bubbling-and-capturing

Also, compare the tests, they're *not* changed, they _expect_ no bubbling for clicks, like in Kirill's spec. The point coordinates are wrong, indeed.

src/anm/events.js
shamansir: Please compare commits and see yourself, if you think I change tests and not the implementation.

spec/mouse.spec.js
shamansir: got it, removed coordinates from `mouseenter`.

>  it might be on a child
so coordinates would be relative to `event.target`.

PR #376, opened 2016-01-18
Add mouse dispatching to Scene, so it acts as a root for mouse events

shamansir:


shamansir:
@kirillk @skavish

shamansir:
https://github.com/Animatron/player/pull/376/files?w=1 (w/o whitespace in diffs)

PR #377, opened 2016-01-19
Enable analytics on production

RobertoUa:
@skavish @shamansir


shamansir:
Is it ok to merge w/o @skavish review?

RobertoUa:
let's wait for @skavish to merge this

skavish:
will it work without analytics, i.e. if we don't have any analytics installed?

RobertoUa:
What do mean by `analytics installed`?
If analytics server is down(or not available) nothing bad will happen.
If use player for adwords the whole `analytics.js` won't be included to the player.

PR #378, opened 2016-01-19
added outline=none to all created canvases

skavish:


skavish:
please review, we need it asap

shamansir:
Actually I did it already: https://github.com/Animatron/player/blame/9fec84b0feab5cc77b40a60da7a1b01dafb56b5b/src/engine/dom-engine.js#L674.

Also, why `newCanvas` is better? it's quite easy to forget space between `new` and `Canvas` and break some staff, also there's `anm.createPlayer` and so on, so it's like common naming system for factory methods.

shamansir:
Uh, got it, it's a wrap for overriding the style. I will cherry-pick the commit I've done instead, if you are not completely against it.

shamansir:
Update: I merged your PR to `master` since I am not sure code there is ok with my version and your version is straight and safe. But my way is safe for `develop` version, so I'll leave it here and then rewrite the `master` with latest version.

PR #379, opened 2016-01-19
added outline=none to all created canvases

skavish:
this is just cherry pick to master, if you think it's fine then just merge it right away


shamansir:
I am not sure `master` version correctly handles `handleEvents` flag, like the fix expects there: https://github.com/Animatron/player/commit/c07acaca484c0576702df83d5f737386dc64c3c1, so I will merge this one and then rewrite it with new version.

PR #383, opened 2016-01-26
Fix passing interactivity events inside groups

shamansir:


shamansir:
@kirillk @skavish pls review

src/anm/events.js
```diff
@@ -164,28 +164,31 @@ MouseEventsSupport.prototype.dispatch = function(event) {
...
-        if (!dispatchedChild) {
-            if (localEvent.type === 'mousemove') {
-                this.processMove(localEvent); // fire mouseenter/mouseexit if required
-            }
-            if ((localEvent.type === 'mouseclick') || (localEvent.type === 'mousedown')) {
-                state.pressedNode = this.owner;
-            }
-            this.owner.fire(localEvent.type, localEvent);
+    // here and below localEvent has properties `.x` and `.y`, so duck typing works
+    dispatchedByOwner = owner.inBounds && owner.inBounds(localEvent) && owner.inside && owner.inside(localEvent);
```
skavish: is it real case when owner does not have .inBounds or .inside?
and I believe if the owner does have .inBounds it will also have .inside, so check for .inside is not necessary

skavish: ok, I see the comment below. then why not add those to scene?

skavish: .inBounds checks if the event is in bounds, what does .inside checks when the owner has children?

skavish: usually checks like that are done this way:
i.e. inBounds is internal optimization of inside. for example for image inBounds is not needed at all, it's the same as inside. in fact I suggest to do it first without inBounds at all.

  primitive.inside(p) {
     return inBounds(p) && insidethiselement(p);
  }

  composite.inside(p) {
     if (inBounds(p)) {
        children.each { child => if (child.inside(p)) return true; }
     }
     return false;
  }

shamansir:
@skavish in this case we should iterate only through visible children and adapt the point (or event) to child matrix before passing it to child `inBounds` test. Also (uh) iterate in reverse, for optimisation (to find and chec top element first). On the other hand all if that could be inside `inBounds` and return the matched child (to pass event further)

skavish:
yes, we need to iterate only through visible children and the point should be transformed into the child's coordinates.
iterate from the top of course.

src/anm/events.js
```diff
@@ -164,28 +164,31 @@ MouseEventsSupport.prototype.dispatch = function(event) {
...
+    // scenes have no .inside or .inBounds methods
```
shamansir: @skavish

> is it real case when owner does not have .inBounds or .inside? 

there should be `.isScene` check there, yes, I'll fix that. 

> then why not add those to scene?

I don't see scene as a visual element which may say "point is inside me", these methods were there before, but while they always should return `true` by logic (if Scene is visible), for this dispatching to work, Scene (as a root element) should return `false`.

skavish: of course scene is visual element. 
- it can paint itself
- it receives events
- it has bounds (0,0,width,height)
- it has children
- it has background
- we can apply transform to scene (camera)
- we can test events whether they are inside scene or not (yes, currently scene gets only events which canvas gets, i.e. all, but still)

in fact scene is almost the same as composite element (clip)

skavish: you can simplify your model if you start treating scene the same as composite element (plus some extra stuff like background)

src/anm/events.js
shamansir: @skavish following [this conversation](https://github.com/Animatron/player/pull/383#issuecomment-175063710):

Actually, in case of mouse events, we need to check every element separately, I mean check if point belongs to the child _and this child dispatched the event_ (is that right, @kirillk?)  and continue iteration only if _both_ conditions are true. I [tried to start this way](https://github.com/Animatron/player/commit/7b3c09fca0b505a36d2419e69980cdeac225dcbc), with `inBounds` iterating through children but haven't found the good way to fit it in dispatching, the way you describe it, it needs to have some callback passed to `.inside` so it may stop iteration if child dispatched the event.

skavish: that's technicality, i.e. implementation detail.

src/anm/events.js
shamansir: Ok, I agree here. At some point we may get back to this, but it touches a lot of things now, so I'd like to postpone it.

src/anm/events.js
kirillk: everything is in the spec and reference implementation: https://dl.dropboxusercontent.com/s/rpnd9ez5thb6phc/2016-01-26%20at%2012.47%20PM.png

I don't know why on earth we're discussing this, having discussed it many times + I spent time clearing it and make the spec

shamansir: > for example for image inBounds is not needed at all

for image inBounds checks if image is rotated and point belongs to its width/height

I changed this line and now it looks a bit better:

```javascript
dispatchedByOwner = !owner.isScene && owner.inside(localEvent);
```

`.inside` checks bounds first: https://github.com/Animatron/player/blob/develop/src/anm/animation/element.js#L1977, but does no checks for children, the reasoning for that (in case of dispatching events) I described above.

I'd like to merge it to see if it works "in real life" and if it's ok, fix code issues next week.

shamansir: @kirillk there's no matrices in spec, no curves paths in spec, it's just an ideal-case example, I already moved everything to the code what was in this spec.

shamansir: @kirillk Also, in spec, there's no iteration through children in `.contains` method, same way it's done in PR, so seems your question "why on earth" was directed not to me?

kirillk: matrices, curves are not relevant to mouse event processing at all

the spec contains the “transfromToChild” and “contains” method that cover
it all
it’s a spec, not an actual implementation
if I’d covered all of that I would have implemented it myself

kirillk: You asked me a question about checking bounds and dispatching to a child. I
answered. Yes, this was directed to you. We’ve discussed that so many times
and we have the spec that right about that.

shamansir: `contains` is relevant both for matrices and curves in real-life case, like we have here, and that's why writing a spec takes two days and implementing it in real life takes more. `transformToChild` === `element.adapt`, `contains` === `element.inside` (sorry they were named differently, but it happened before the spec), both are already here. In your spec `contains` is *not* checking children, so I suppose it's the answer to Dima -> it should not iterate through children.

PR #384, opened 2016-02-02
Interactivity mouse: Adapting new specification

shamansir:


shamansir:
Tests yet failing.

shamansir:
@skavish @kirillk all tests pass now (let's wait for Travis/TC just in case), ok to review and merge.

skavish:
I am not sure what is going on with travis build. is it failed or not?

src/anm/animation/scene.js
```diff
@@ -101,6 +105,20 @@ Scene.prototype.reverseEachVisible = function(visitor, data) {
     }
 };
 
+Scene.prototype.findDeepestChildAt = function(local_pt) {
```
skavish: Scene should be made element (at some point, not now) at least to avoid this code dups

src/anm/graphics/text.js
```diff
@@ -171,7 +171,7 @@ Text.prototype.bounds = function() {
  * @param {Number} pt.y
  * @return {Boolean} is point inside
  */
-Text.prototype.inside = function(pt) {
+Text.prototype.contains = function(pt) {
```
skavish: well, this optimization implies that "contains" always go after "inBounds". is it really true?

src/anm/media/video.js
```diff
@@ -158,7 +158,7 @@ Video.prototype.bounds = function() {
  * @param {Number} pt.y
  * @return {Boolean} is point inside
  */
-Video.prototype.inside = function(pt) {
+Video.prototype.contains = function(pt) {
     return true; // if point is inside of the bounds, point is considered to be
```
skavish: same here

src/anm/graphics/text.js
shamansir: https://github.com/Animatron/player/pull/384/files#diff-a072a28291a97d7ac72e402357d42fc6R1979

It is, though the fact it's always named `contains` probably requires `Text`, `Video` etc. to also have `inBounds` method, just to make it similar everywhere.

PR #387, opened 2016-02-20
Interactivity: Messages

shamansir:
```javascript
// Global:

animation.fireMessage('foo'); 
// which is a shortcut to:
animation.fire('message', 'foo');

animation.fireMessageAt(2.0, 'foo'); 
// which is a shortcut to:
animation.timeline.at(2.0, function() { animation.fireMessage('foo'); }); 

animation.onMessage('foo', function() { ... }); 
// which is a shortcut to:
animation.on('message', function(msg) { if (msg === 'foo') ... }); 

// Local:

element.fireMessage('foo'); 
// which is a shortcut to:
element.fire('message', 'foo');

element.fireMessageAt(2.0, 'foo'); 
// which is a shortcut to:
element.timeline.at(2.0, function() { element.fireMessage('foo') }); 

element.onMessage('foo', function() { ... }); 
// which is a shortcut to:
element.on('message', function(msg) { if (msg === 'foo') ... }); 
```

### DIFF
--- a/spec/time.spec.js
+++ b/spec/time.spec.js
@@ -775,6 +775,7 @@ describe('time', function() {
 
             it('fires a message at requested time', function() {
                 var anim = new anm.Animation();
+                anim.setDuration(10);
 
                 var fooSpy = jasmine.createSpy('foo');
                 anim.onMessage('foo', fooSpy);
@@ -866,10 +867,10 @@ describe('time', function() {
                 var anim = new anm.Animation();
                 var root = new anm.Element();
                 anim.add(root);
+                anim.setDuration(10);
 
                 var fooSpy = jasmine.createSpy('foo');
                 root.onMessage('foo', fooSpy);
-
 
                 // anim: 0-----1-----2-----3-----4-----5-----6-
                 // root: 0-----1---#foo#---3-----4-----5-----6-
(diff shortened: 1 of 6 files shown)


shamansir:
@kirillk @skavish

kirillk:
Sending a message on element as well as receiving that on an element doesn't make sense, that all should be done on the bus object.

Making a separate event handler to send a message in a time event is over engineering as well.

So that should be that simple as done in the editor code generation, there is no more sense behind this. It's literally a few lines of code.

shamansir:
> Sending a message on element as well as receiving that on an element doesn't make sense, that all should be done on the bus object.

Agree, at least in UI it's not needed for the moment. For API, it could be useful.

> Making a separate event handler to send a message in a time event is over engineering as well.

Why not? Start several animations at one moment of time is quite useful. 

> So that should be that simple as done in the editor code generation, there is no more sense behind this. It's literally a few lines of code.

It's definitely same easy, it utilises the event system (say _event bus_) which is there for a long time and based on the same principle. Have you looked trough the code? I've looked through yours.

My point is since we use Player API in Interactivity (and still have plans exposing it), Messages should also be there. And (especially in comparison to time events) it costs almost nothing to implement them, we discuss it more than it took to do it.

PR #388, opened 2016-02-29
Audio fixes

shamansir:
Fixes for most of the present audio problems (especially Master Audio and stopping audio when animation played to the end).


shamansir:
@skavish plz review.

PR #391, opened 2016-03-15
don't add audio extension if already present

RobertoUa:
@shamansir @skavish 
not sure, but it's probably better to check like this `\.(mp3|ogg|wav)$`

### DIFF
--- a/src/anm/media/audio.js
+++ b/src/anm/media/audio.js
@@ -47,7 +47,7 @@ var audioContext = getAudioContext();
  * @class anm.Audio
  */
 function Audio(url) {
-    this.url = url + audioExt;
+    this.url = /\.\w+$/i.test(url) ? url : url + audioExt;
     this.ready = false;
     this.active = false;
     this.playing = false;


src/anm/media/audio.js
```diff
@@ -47,7 +47,7 @@ var audioContext = getAudioContext();
  * @class anm.Audio
  */
 function Audio(url) {
-    this.url = url + audioExt;
+    this.url = /\.\w+$/i.test(url) ? url : url + audioExt;
```
shamansir: @RobertoUa Please note that when we add audio source, `audioType` is also required. Here it is passed to a function: https://github.com/Animatron/player/pull/392/files#diff-4e959658da38c66435f3dd42b062b504L183 , and here it is constructed: https://github.com/Animatron/player/pull/391/files#diff-4e959658da38c66435f3dd42b062b504R14.

skavish: i suggest to use type depending on the media

shamansir: @skavish As I recall, there was a high chance that two sources (both `ogg` and `mp3`) exist on server and that's why we used audio URL w/o extension — this way we have chosen what to play by checking if user's browser supports this format, not by dictating it to user.

RobertoUa: what if we have one source of audio? We have such case now.

RobertoUa: anyway, player should support url with and without extension.

skavish: yes, you are right. but in cases when there is exactly one source with a known extension it is not going to work. and looks like we will have cases like that. now for this project Vova is working on

shamansir: JFYI, for video we have a list of supported formats and URLs: e5e65856.

skavish: yes, we will soon start exporting audio the same way (not really soon, but soon)

shamansir: got it, thanks. then indeed, _only in this case_ (only when extension is already present), we should load audio type from media or else it will be `audio/ogg` for `.mp3` and s.o.

> we will soon start exporting audio the same way (not really soon, but soon)

good!

shamansir: @RobertoUa So, will this one be fixed? It's a potential issue, I suppose.

PR #393, opened 2016-03-18
Check audio type based on extension

RobertoUa:
@shamansir


shamansir:
Super-cool! Thanks a lot!

